PR #8, opened 2017-04-11
Parser interface and NullParser

gfr10598:
This introduces the Parser interface, and a NullParser implementation.  Real and Pseudo parser implementations should be based on this code.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/8)

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -1,0 +1,48 @@
+// The parser package defines the Parser interface and implementations for the different
+// test types, NDT, Paris Traceroute, and SideStream.
+package parser
+
+import (
+	"cloud.google.com/go/bigquery"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+//=====================================================================================
+//                       Parser Interface and implementations
+//=====================================================================================
+type Parser interface {
+	// fn - Name of test file
+	// table - biq query table name (for error logging only)
+	// test - binary test data
+	HandleTest(fn string, table string, test []byte) (bigquery.ValueSaver, error)
+}
+
+type NullParser struct {
+	Parser
+}
+
+func (np *NullParser) HandleTest(fn string, table string, test []byte) (bigquery.ValueSaver, error) {
+	test_count.With(prometheus.Labels{"table": table}).Inc()
+	return nil, nil
+}
+
+//=====================================================================================
+//                       Prometheus Monitoring
+//=====================================================================================
+
+var (
+	test_count = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "etl_parser_test_count",
+		Help: "Number of tests processed.",
+	}, []string{"table"})
+
+	failure_count = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "etl_parser_failure_count",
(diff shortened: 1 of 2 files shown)


stephen-soltesz:
:lgtm:

---

Reviewed 1 of 10 files at r3.
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[parser/parser.go, line 25 at r4](https://reviewable.io:443/reviews/m-lab/etl/8#-KhTR9dVNJGSAkH-3UBY:-KhTR9dVNJGSAkH-3UBZ:booi41d) ([raw file](https://github.com/m-lab/etl/blob/c92b1714f4e571ff1019c58efcea573990ea2150/parser/parser.go#L25)):*
> ```Go
> 
> func (np *NullParser) HandleTest(fn string, table string, test []byte) (bigquery.ValueSaver, error) {
> 	test_count.With(prometheus.Labels{"table": table}).Inc()
> ```

I like the idea of us being able to define "null" HandleTest implementation that wraps a real implementation (analogous to how `http.Handler` functions can wrap one another) so that we can transparently count calls, and execution times.

---

*[parser/parser.go, line 34 at r4](https://reviewable.io:443/reviews/m-lab/etl/8#-KhTRdQlynEAAnNsQwnB:-KhTRdQmLh5HmrjCbe_r:b-nwfgq9) ([raw file](https://github.com/m-lab/etl/blob/c92b1714f4e571ff1019c58efcea573990ea2150/parser/parser.go#L34)):*
> ```Go
> 
> var (
> 	test_count = prometheus.NewCounterVec(prometheus.CounterOpts{
> ```

I imagined placing these in a separate `metrics` package so that all metrics are defined in one place.  This may make it easier to enforce consistent naming and documentation conventions over time.

What are the other trade offs to consider?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/8#-:-KhTSlrqKXpnXna05Fbm:bnfp4nl)*

gfr10598:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[parser/parser.go, line 34 at r4](https://reviewable.io:443/reviews/m-lab/etl/8#-KhTRdQlynEAAnNsQwnB:-KhTVc9DaZPVDG9__sSo:bf83vdn) ([raw file](https://github.com/m-lab/etl/blob/c92b1714f4e571ff1019c58efcea573990ea2150/parser/parser.go#L34)):*

I imagined placing these in a separate `metrics` package so that all metrics are defined in one place.  This may make it easier to enforce consistent naming and documentation conventions over time.

What are the other trade offs to consider?
</blockquote></details>

I think the main tradeoff is that anyone adding a new metric would have to modify code in another package.

I'm guessing we will converge on a combination, where some of the metrics are global and located in a common place, and other metrics are of narrower interest, and perhaps better served with encapsulation rather than global scope.

Of course, we have just created almost identical metrics in our two PRs.  I'm guessing your's will survive in this case.

BTW, I like your comments with Usage examples (in other PR)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/8)*

PR #16, opened 2017-04-12
Alpha BQ Inserter, and integration into Alpha Task

gfr10598:
Introduces a simple BQ inserter object.

Thanks to Ya for proof of concept that made this easier.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/16)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,9 +31,11 @@ cache:
   directories:
     - "$HOME/google-cloud-sdk/"
 
-
 script:
-- go test -v ./...
+# Currently skipping bq tests, because they depend on bigquery, and there is
+# no emulator
+- go test -v github.com/m-lab/etl/parser
+- go test -v github.com/m-lab/etl/task
 - cd $TRAVIS_BUILD_DIR/cmd/etl_worker && go build
 - $TRAVIS_BUILD_DIR/travis/install_gcloud.sh
 
--- a/bq/bq_test.go
+++ b/bq/bq_test.go
@@ -7,7 +7,17 @@ import (
 	"time"
 )
 
-// Bad - has side effects.
+// Item represents a row item.
+type Item struct {
+	Name   string
+	Count  int
+	Foobar int `json:"foobar"`
+}
+
+// NB: This test has side effects and depends on BigQuery service and
+// test table.
+// Do not run this test from travis.
+// TODO - use emulator when available.
 func TestInsert(t *testing.T) {
 	tag := "new"
 	items := []*Item{
--- a/bq/bq_test.go
+++ b/bq/bq_test.go
@@ -1,0 +1,28 @@
+package bq
+
+import (
+	"fmt"
+	"os"
+	"testing"
+	"time"
+)
+
+// Bad - has side effects.
+func TestInsert(t *testing.T) {
+	tag := "new"
+	items := []*Item{
+		// Each item implements the ValueSaver interface.
+		{Name: tag + "_x0", Count: 17, Foobar: 44},
+		{Name: tag + "_x1", Count: 12, Foobar: 44},
+	}
+
+	in, err := NewInserter("mlab-sandbox", "mlab_sandbox", "test2")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		return
+	}
+
+	if err = in.InsertRows(items, 10*time.Second); err != nil {
+		fmt.Fprintf(os.Stderr, "failed to insert rows: %v\n", err)
+	}
+}
--- a/bq/insert.go
+++ b/bq/insert.go
@@ -1,21 +1,24 @@
 package bq
 
 import (
-	//"log"
 	"time"
 
 	"cloud.google.com/go/bigquery"
 	"golang.org/x/net/context"
 )
 
-// An Inserter
-type Inserter struct {
+// An Inserter provides the InsertRows function.
+type Inserter interface {
+	InsertRows(data interface{}, timeout time.Duration) error
+}
+
+type BQInserter struct {
+	Inserter
 	client   *bigquery.Client
 	uploader *bigquery.Uploader
 }
 
-func NewInserter(project string, dataset string, table string) (*Inserter, error) {
-
+func NewInserter(project string, dataset string, table string) (Inserter, error) {
 	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
 	// Heavyweight!
 	client, err := bigquery.NewClient(ctx, project)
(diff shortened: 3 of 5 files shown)


stephen-soltesz:
Reviewed 4 of 5 files at r1.
Review status: 4 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[bq/insert.go, line 31 at r1](https://reviewable.io:443/reviews/m-lab/etl/16#-KhYyYu5VBX2klx0EqR0:-KhYsOL5ZoW6JInNznFU:b-vmf6hu) ([raw file](https://github.com/m-lab/etl/blob/7f7ad6cda7989ec7aec92d3ef4f575ebee51ca64/bq/insert.go#L31)):*
> ```Go
> type BQInserter struct {
> 	Inserter
> 	client   *bigquery.Client
> ```

Did you decide against anonymous fields here to keep them private?

---

*[bq/insert.go, line 32 at r1](https://reviewable.io:443/reviews/m-lab/etl/16#-KhYtOGABV_T3_EPkfY2:-KhYtOGABV_T3_EPkfY3:b-1fa8ll) ([raw file](https://github.com/m-lab/etl/blob/7f7ad6cda7989ec7aec92d3ef4f575ebee51ca64/bq/insert.go#L32)):*
> ```Go
> 	Inserter
> 	client   *bigquery.Client
> 	uploader *bigquery.Uploader
> ```

When the number of methods needed on a public API like `bigquery.Uploader` is relatively few, I would like to advocate for using private interfaces so we can write unit tests that do not require external services (perhaps in addition to those that do for an integration test). This gets cumbersome when the number of methods on the public interface becomes large, but for moderate or small sized interfaces, I think the benefits will accrue to us over time by creating simpler, more comprehensive, and faster unit tests as compared to live integration-as-unit tests.

An example of this approach in epoxy:
https://github.com/m-lab/epoxy/blob/dev/storage/datastore.go#L30

---

*[bq/insert.go, line 38 at r1](https://reviewable.io:443/reviews/m-lab/etl/16#-KhYtus4LM9iPte-Qrnr:-KhYtus4LM9iPte-Qrns:bk8uz9t) ([raw file](https://github.com/m-lab/etl/blob/7f7ad6cda7989ec7aec92d3ef4f575ebee51ca64/bq/insert.go#L38)):*
> ```Go
> 	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
> 	// Heavyweight!
> 	client, err := bigquery.NewClient(ctx, project)
> ```

What do you think of using the object injection pattern here, where the caller defines the client object and passes it to the "NewInserter" function? My understanding is that object injection is preferable for hard-to-test logic because we have the option of providing our own implementation during tests.

In epoxy, we arrived at this pattern with the datastore interface. I think it would be helpful here.
https://github.com/m-lab/epoxy/blob/dev/cmd/epoxy_boot_server/main.go#L132

---

*[task/task_test.go, line 64 at r1](https://reviewable.io:443/reviews/m-lab/etl/16#-KhYzgLA40RCm51NC5v7:-KhYzgLA40RCm51NC5v8:b5kiwpr) ([raw file](https://github.com/m-lab/etl/blob/7f7ad6cda7989ec7aec92d3ef4f575ebee51ca64/task/task_test.go#L64)):*
> ```Go
> 
> 	var prsr TestParser
> 	in := bq.NullInserter{}
> ```

Object injection for the win! :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/16)*

gfr10598:
Review status: 4 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[bq/insert.go, line 31 at r1](https://reviewable.io:443/reviews/m-lab/etl/16#-KhYyYu5VBX2klx0EqR0:-KhZ0NqJqqwxhXnoldT6:bz24tbc) ([raw file](https://github.com/m-lab/etl/blob/7f7ad6cda7989ec7aec92d3ef4f575ebee51ca64/bq/insert.go#L31)):*

Did you decide against anonymous fields here to keep them private?
</blockquote></details>

Basically correct.  There is no need to export the interfaces of these objects.  Just need them internally.

---

*[bq/insert.go, line 32 at r1](https://reviewable.io:443/reviews/m-lab/etl/16#-KhYtOGABV_T3_EPkfY2:-KhZ0eRtHVtHNNLT2v0Z:b-c37ddb) ([raw file](https://github.com/m-lab/etl/blob/7f7ad6cda7989ec7aec92d3ef4f575ebee51ca64/bq/insert.go#L32)):*

When the number of methods needed on a public API like `bigquery.Uploader` is relatively few, I would like to advocate for using private interfaces so we can write unit tests that do not require external services (perhaps in addition to those that do for an integration test). This gets cumbersome when the number of methods on the public interface becomes large, but for moderate or small sized interfaces, I think the benefits will accrue to us over time by creating simpler, more comprehensive, and faster unit tests as compared to live integration-as-unit tests.

An example of this approach in epoxy:
https://github.com/m-lab/epoxy/blob/dev/storage/datastore.go#L30
</blockquote></details>

Excellent point, and I've suggested the same to Ya.  I think in this case the value would be minimal, and we get most of what we need from injecting at the level of the Inserter interface.  Adding a TODO to consider alternatives, though.

FYI Uploader has a single Put() function, so mocking that would be pretty easy, but I'm more interested in encapsulating the Client, which is a precursor to the Uploader.

---

*[bq/insert.go, line 38 at r1](https://reviewable.io:443/reviews/m-lab/etl/16#-KhYtus4LM9iPte-Qrnr:-KhZ4hvuZT_whg8LGdTR:bsxn7wb) ([raw file](https://github.com/m-lab/etl/blob/7f7ad6cda7989ec7aec92d3ef4f575ebee51ca64/bq/insert.go#L38)):*

What do you think of using the object injection pattern here, where the caller defines the client object and passes it to the "NewInserter" function? My understanding is that object injection is preferable for hard-to-test logic because we have the option of providing our own implementation during tests.

In epoxy, we arrived at this pattern with the datastore interface. I think it would be helpful here.
https://github.com/m-lab/epoxy/blob/dev/cmd/epoxy_boot_server/main.go#L132
</blockquote></details>

See previous reply.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/16)*

stephen-soltesz:
:lgtm:

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/16#-:-KhZ7gyBV_iHNd-R-udN:bnfp4nl)*

PR #17, opened 2017-04-12
Utility functions for gcs tar file

gfr10598:
Not ready for review, but if  you have any helpful suggestions they would be welcome

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/17)

### DIFF
--- a/storage/gcs.go
+++ b/storage/gcs.go
@@ -1,0 +1,97 @@
+package storage
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"golang.org/x/net/context"
+	"golang.org/x/oauth2/google"
+	storage "google.golang.org/api/storage/v1"
+	"net/http"
+	"strings"
+	"time"
+)
+
+type TarReaderCloser struct {
+	*tar.Reader
+	// Should this be a slice?
+	closer func() error
+}
+
+func (trc *TarReaderCloser) Close() {
+	trc.closer()
+}
+
+// TODO - is gensupport.URLParams useful?
+// Create a tar.Reader suitable for injecting into Task.
+// When do we close the response body?
+func newTarReader(uri string) (*TarReaderCloser, error) {
+
+	client, err := getClient()
+	if err != nil {
+		return nil, err
+	}
+	obj, err := getObject(client, uri, 60*time.Second)
+	if err != nil {
+		return nil, err
+	}
+
+	// Wrap with a tar.Reader that also has a response closer.
+
+	// Is it a tar or tgz file?
+	rdr := obj.Body
+	closer := obj.Body.Close
+
+	if strings.HasSuffix(strings.ToLower(uri), ".gz") {
+		// TODO add unit test
+		rdr, err := gzip.NewReader(obj.Body)
+		if err != nil {
+			closer()
+			return nil, err
+		}
+		c := func() error {
+			defer closer()
+			rdr.Close()
+			// TODO handle errors?
+			return nil
+		}
+		closer = c
+	}
+	tarReader := tar.NewReader(rdr)
+	// No closer needed for tar.Reader
+
+	return &TarReaderCloser{tarReader, closer}, nil
+}
+
+func getClient() (*http.Client, error) {
+	// TODO - is this the scope we want?
+	scope := storage.DevstorageFullControlScope
+	// Use a short timeout, so we get an error quickly if there is a problem.
+	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
+	client, err := google.DefaultClient(ctx, scope)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
+}
+
+// Caller responsible for closing response body.
+func getObject(client *http.Client, uri string, timeout time.Duration) (*http.Response, error) {
+	// Lightweight, error only if client is nil.
+	service, err := storage.New(client)
+	if err != nil {
+		return nil, err
+	}
+
+	// How do we parse the uri to get the bucket and object names?
+	call := service.Objects.Get("", "") // sourceBucket, objectName)
+	ctx, _ := context.WithTimeout(context.Background(), timeout)
+	call = call.Context(ctx)
+
+	// Heavyweight.
+	// Doesn't look like any googleapi.CallOptions are useful here.
+	contentResponse, err := call.Download()
+	if err != nil {
+		return nil, err
+	}
+	return contentResponse, err
+}
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -1,3 +1,9 @@
+// GCS related utility functions to fetch and wrap objects with tar.Reader.
+//
+// Testing:
+//   This has been manually tested, but test automation is probably not
+//   worthwhile until there is an emulator for GCS.
+
 package storage
 
 import (
@@ -23,9 +29,12 @@ func (trc *TarReaderCloser) Close() {
 	trc.closer()
 }
 
-// TODO - is gensupport.URLParams useful?
 // Create a tar.Reader suitable for injecting into Task.
-func newGCSTarReader(uri string) (*TarReaderCloser, error) {
+// Caller is responsible for calling Close on the returned object.
+//
+// uri should be of form gs://bucket/filename.tar or gs://bucket/filename.tgz
+// FYI Using a persistent client saves about 80 msec, and 220 allocs, totalling 70kB.
+func NewGCSTarReader(client *http.Client, uri string) (*TarReaderCloser, error) {
 	// For now only handle gcs paths.
 	if !strings.HasPrefix(uri, "gs://") {
 		return nil, errors.New("invalid file path: " + uri)
@@ -37,11 +46,10 @@ func newGCSTarReader(uri string) (*TarReaderCloser, error) {
 	bucket := parts[2]
 	fn := parts[3]
 
-	// TODO reuse client.
-	client, err := getClient()
-	if err != nil {
-		return nil, err
+	if !(strings.HasSuffix(fn, ".tgz") || strings.HasSuffix(fn, ".tar")) {
+		return nil, errors.New("not tar or tgz: " + uri)
 	}
+
 	obj, err := getObject(client, bucket, fn, 60*time.Second)
 	if err != nil {
 		return nil, err
@@ -52,16 +60,17 @@ func newGCSTarReader(uri string) (*TarReaderCloser, error) {
 	closer := obj.Body.Close
 
 	// Is it a tar or tgz file?
-	if strings.HasSuffix(strings.ToLower(fn), ".gz") {
+	if strings.HasSuffix(strings.ToLower(fn), ".tgz") {
 		// TODO add unit test
 		rdr, err := gzip.NewReader(obj.Body)
 		if err != nil {
 			closer()
 			return nil, err
 		}
+		tmp := closer
 		c := func() error {
-			defer closer()
 			rdr.Close()
+			tmp()
 			// TODO handle errors?
 			return nil
 		}
@@ -73,9 +82,18 @@ func newGCSTarReader(uri string) (*TarReaderCloser, error) {
 	return &TarReaderCloser{tarReader, closer}, nil
 }
 
-func getClient() (*http.Client, error) {
+//---------------------------------------------------------------------------------
+//          Local functions
+//---------------------------------------------------------------------------------
+
+// Create a storage reader client.
+func getStorageClient(write_access bool) (*http.Client, error) {
 	// TODO - is this the scope we want?
-	scope := storage.DevstorageFullControlScope
+	var scope string
+	if scope = storage.DevstorageReadOnlyScope; write_access {
+		scope = storage.DevstorageReadWriteScope
+	}
+
 	// Use a short timeout, so we get an error quickly if there is a problem.
 	ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
 	client, err := google.DefaultClient(ctx, scope)
@@ -86,7 +104,7 @@ func getClient() (*http.Client, error) {
 	return client, nil
 }
 
-// Caller responsible for closing response body.
+// Caller is responsible for closing response body.
 func getObject(client *http.Client, bucket string, fn string, timeout time.Duration) (*http.Response, error) {
 	// Lightweight, error only if client is nil.
 	service, err := storage.New(client)
@@ -94,6 +112,7 @@ func getObject(client *http.Client, bucket string, fn string, timeout time.Durat
 		return nil, err
 	}
 
+	// Lightweight - only setting up the local object.
 	call := service.Objects.Get(bucket, fn)
 	ctx, _ := context.WithTimeout(context.Background(), timeout)
 	call = call.Context(ctx)
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -83,7 +83,6 @@ func NewGCSTarReader(client *http.Client, uri string) (*TarReaderCloser, error) 
 	// Is it a tar.gz or tgz file?
 	if strings.HasSuffix(strings.ToLower(fn), "gz") {
 		// TODO add unit test
-		var err error
 		// NB: This must not be :=, or it creates local rdr.
 		rdr, err = gzip.NewReader(obj.Body)
 		if err != nil {
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -14,20 +14,28 @@ import (
 	"golang.org/x/net/context"
 	"golang.org/x/oauth2/google"
 	storage "google.golang.org/api/storage/v1"
+	"io"
 	"net/http"
 	"strings"
 	"time"
 )
 
+type TarReader interface {
+	Next() (*tar.Header, error)
+	Read(b []byte) (int, error)
+}
+
 type TarReaderCloser struct {
-	*tar.Reader
-	// Should this be a slice?
+	TarReader
+	io.Closer
+}
+
+type C struct {
 	closer func() error
 }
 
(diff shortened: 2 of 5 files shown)


stephen-soltesz:
Some initial thoughts.

I'm still wondering how we can define a better interface around the tar reader / gzip reader / body closer.

---

Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions.

---

*[storage/storage.go, line 22 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khd-sySxdafVOexnTY4:-Khd-sySxdafVOexnTY5:b7jw5ik) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L22)):*
> ```Go
> )
> 
> type TarReaderCloser struct {
> ```

I'm wondering if we could define our own interface around an implementation using the tar.Reader. e.g. Something like:

```
type GCSTarReaderCloser interface {
    func (tr *Reader) Next() (*Header, error)
    func (tr *Reader) Read(b []byte) (int, error)
    io.Closer
}
```

And writing tiny wrapper functions for calling the underlying `tar` object.

---

*[storage/storage.go, line 25 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-KhcyG6es91aOkBXa1D_:-KhcyG6es91aOkBXa1Da:b-ihpp1d) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L25)):*
> ```Go
> 	*tar.Reader
> 	// Should this be a slice?
> 	closer func() error
> ```

What do you think of using a standard `Closer` interface? https://golang.org/pkg/io/#Closer

---

*[storage/storage.go, line 39 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-KhcquAUgS9NFm8ikq8J:-KhcquAUgS9NFm8ikq8K:b30j2bq) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L39)):*
> ```Go
> func NewGCSTarReader(client *http.Client, uri string) (*TarReaderCloser, error) {
> 	// For now only handle gcs paths.
> 	if !strings.HasPrefix(uri, "gs://") {
> ```

It's too bad the storage API does not provide a mechanism for parsing these URIs.

The golang `net/url` package can parse generic URL schemes. https://golang.org/pkg/net/url/#URL

The simpler parsing might be worth tolerating calling the bucketname "Host" briefly.

---

*[storage/storage.go, line 49 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-KhcoXy7YAuIQ0qsaGcn:-KhcoXy8Y1_JjKodOobM:b-ah89fp) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L49)):*
> ```Go
> 	fn := parts[3]
> 
> 	if !(strings.HasSuffix(fn, ".tgz") || strings.HasSuffix(fn, ".tar")) {
> ```

Just a clarifying question, we have no ".tar.gz" files?

---

*[storage/storage.go, line 58 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khcx87E8PSH9Pnk5eS-:-Khcx87E8PSH9Pnk5eS0:b-6paiat) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L58)):*
> ```Go
> 	}
> 
> 	// Wrap with a tar.Reader that also has a response closer.
> ```

I don't have a counter example, but some part of me feels like the "Closer" setup below could be simpler somehow.

---

*[storage/storage.go, line 67 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-KhctxCyH9EXoq-8tH9e:-KhctxCyH9EXoq-8tH9f:b-e7vstq) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L67)):*
> ```Go
> 		rdr, err := gzip.NewReader(obj.Body)
> 		if err != nil {
> 			closer()
> ```

Can we use `obj.Body.Close` directly here?

---

*[storage/storage.go, line 73 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khcto4z3NL3YE_NvZ-E:-Khcto4z3NL3YE_NvZ-F:b-5inc2) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L73)):*
> ```Go
> 		c := func() error {
> 			rdr.Close()
> 			tmp()
> ```

Do we need `tmp`? Could we use `obj.Body.Close` directly here?

---

*[storage/storage.go, line 90 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-KhcusuRq2w0oL5ZPjak:-KhcusuRq2w0oL5ZPjal:b-usvy5i) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L90)):*
> ```Go
> 
> // Create a storage reader client.
> func getStorageClient(write_access bool) (*http.Client, error) {
> ```

style: go discourages snake-case variables names in favor of camel-case. Here `writeAccess` or `withWrite` (or similar) would be preferable.  https://golang.org/doc/effective_go.html#mixed-caps

---

*[storage/storage.go, line 93 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khcw8sdnhTnsXjCHvHv:-Khcw8sdnhTnsXjCHvHw:bvwcesd) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L93)):*
> ```Go
> 	// TODO - is this the scope we want?
> 	var scope string
> 	if scope = storage.DevstorageReadOnlyScope; write_access {
> ```

The logic implies to my mind that `scope = storage.DevstorageReadOnlyScope` is something that could fail, but of course it can't.

I think it would be preferable to separate these two statements.
```
var scope string
scope = storage.DevstorageReadOnlyScope
if write_access {
    scope = storage.DevstorageReadWriteScope
}
```

Or highlight that they're mutually exclusive:
```
var scope string
if write_access {
    scope = storage.DevstorageReadWriteScope
} else {
    scope = storage.DevstorageReadOnlyScope
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/17#-:-Khd2oLwHSgm7ejwEjz7:b-x6ayin)*

gfr10598:
Review status: 0 of 4 files reviewed at latest revision, 9 unresolved discussions.

---

*[storage/storage.go, line 22 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khd-sySxdafVOexnTY4:-KhdE8u62rCCSoutIGuE:b-k0gefk) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L22)):*

I'm wondering if we could define our own interface around an implementation using the tar.Reader. e.g. Something like:

```
type GCSTarReaderCloser interface {
    func (tr *Reader) Next() (*Header, error)
    func (tr *Reader) Read(b []byte) (int, error)
    io.Closer
}
```

And writing tiny wrapper functions for calling the underlying `tar` object.
</blockquote></details>

I was starting to do just that. 

---

*[storage/storage.go, line 25 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-KhcyG6es91aOkBXa1D_:-KhdGlmG7BYpqBYbYla6:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L25)):*

What do you think of using a standard `Closer` interface? https://golang.org/pkg/io/#Closer
</blockquote></details>

Done.

---

*[storage/storage.go, line 39 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-KhcquAUgS9NFm8ikq8J:-KhdHxZpvAb3e0YS3aYN:bvv3a9m) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L39)):*

It's too bad the storage API does not provide a mechanism for parsing these URIs.

The golang `net/url` package can parse generic URL schemes. https://golang.org/pkg/net/url/#URL

The simpler parsing might be worth tolerating calling the bucketname "Host" briefly.
</blockquote></details>

Prefer to leave as is for now.

---

*[storage/storage.go, line 49 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-KhcoXy7YAuIQ0qsaGcn:-KhdGr_JDj59s8DB8yCX:bt2vr1b) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L49)):*

Just a clarifying question, we have no ".tar.gz" files?
</blockquote></details>

Fixed.

---

*[storage/storage.go, line 58 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khcx87E8PSH9Pnk5eS-:-KhdI1S5HgikgVHn-4Kj:b2nhysk) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L58)):*

I don't have a counter example, but some part of me feels like the "Closer" setup below could be simpler somehow.
</blockquote></details>

ack

---

*[storage/storage.go, line 67 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-KhctxCyH9EXoq-8tH9e:-KhdGxMi0iTU3AG9265V:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L67)):*

Can we use `obj.Body.Close` directly here?
</blockquote></details>

Done.

---

*[storage/storage.go, line 73 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khcto4z3NL3YE_NvZ-E:-KhdGy4f6W1GD5WnRynA:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L73)):*

Do we need `tmp`? Could we use `obj.Body.Close` directly here?
</blockquote></details>

Done.

---

*[storage/storage.go, line 90 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-KhcusuRq2w0oL5ZPjak:-KhdHpipYPFcq7_2R5e0:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L90)):*

style: go discourages snake-case variables names in favor of camel-case. Here `writeAccess` or `withWrite` (or similar) would be preferable.  https://golang.org/doc/effective_go.html#mixed-caps
</blockquote></details>

Done.

---

*[storage/storage.go, line 93 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khcw8sdnhTnsXjCHvHv:-KhdHr8sTzkUvwzxsJ5h:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L93)):*

The logic implies to my mind that `scope = storage.DevstorageReadOnlyScope` is something that could fail, but of course it can't.

I think it would be preferable to separate these two statements.
```
var scope string
scope = storage.DevstorageReadOnlyScope
if write_access {
    scope = storage.DevstorageReadWriteScope
}
```

Or highlight that they're mutually exclusive:
```
var scope string
if write_access {
    scope = storage.DevstorageReadWriteScope
} else {
    scope = storage.DevstorageReadOnlyScope
}
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/17)*

stephen-soltesz:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[storage/storage.go, line 58 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khcx87E8PSH9Pnk5eS-:-KhdcRR7IjPtxaO8DlNc:b-8pyzxp) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L58)):*

ack
</blockquote></details>

This is closer to what I'm thinking: https://play.golang.org/p/Reuxq1-x0S

Since we need to close the original `obj.Body` as well as the `gzip.Reader` if we use it, we need references to both objects for Close. In this example, we simply keep references to the underlying types and conditionally close them in the receiver's `Close` implementation.

---

*[storage/storage.go, line 44 at r2](https://reviewable.io:443/reviews/m-lab/etl/17#-Khdn7nFcEWlLaym3-LU:-Khdn7nFcEWlLaym3-LV:b46nknz) ([raw file](https://github.com/m-lab/etl/blob/acb032a27c0957032fae971fb046797729c85ab2/storage/storage.go#L44)):*
> ```Go
> //
> // uri should be of form gs://bucket/filename.tar or gs://bucket/filename.tgz
> // FYI Using a persistent client saves about 80 msec, and 220 allocs, totalling 70kB.
> ```

Based on the benchmark, which approach do you recommend?

My bias is towards always using injection and not defining special cases to auto-allocate objects. Convenience methods that bundle up sequences of necessary calls make sense to me too. But, I imagine those being in a separate package that calls `storage`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/17)*

gfr10598:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[storage/storage.go, line 58 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khcx87E8PSH9Pnk5eS-:-Khht75hem38FAK8Fmvc:b-otsl3u) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L58)):*

This is closer to what I'm thinking: https://play.golang.org/p/Reuxq1-x0S

Since we need to close the original `obj.Body` as well as the `gzip.Reader` if we use it, we need references to both objects for Close. In this example, we simply keep references to the underlying types and conditionally close them in the receiver's `Close` implementation.
</blockquote></details>

Ok - did something similar, see what you think.

---

*[storage/storage.go, line 44 at r2](https://reviewable.io:443/reviews/m-lab/etl/17#-Khdn7nFcEWlLaym3-LU:-Khdrf1wfGmmd7Yekb1Z:b-yka53b) ([raw file](https://github.com/m-lab/etl/blob/acb032a27c0957032fae971fb046797729c85ab2/storage/storage.go#L44)):*

Based on the benchmark, which approach do you recommend?

My bias is towards always using injection and not defining special cases to auto-allocate objects. Convenience methods that bundle up sequences of necessary calls make sense to me too. But, I imagine those being in a separate package that calls `storage`.
</blockquote></details>

Passing the client in, or having a preallocated client in a wrapping object!  Saves a round trip to the server.  We may need to check for disconnections or some other rare conditions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/17)*

stephen-soltesz:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[storage/storage.go, line 58 at r1](https://reviewable.io:443/reviews/m-lab/etl/17#-Khcx87E8PSH9Pnk5eS-:-Khi0Ah_R1G9ifz9oilB:b8eau4p) ([raw file](https://github.com/m-lab/etl/blob/8dd6478a8088d5045a806f365f5c17662da63bdd/storage/storage.go#L58)):*

Ok - did something similar, see what you think.
</blockquote></details>

I like it!

---

*[storage/storage.go, line 44 at r2](https://reviewable.io:443/reviews/m-lab/etl/17#-Khdn7nFcEWlLaym3-LU:-Khi-UPmKU8uzTYsyVcC:bb6mt9b) ([raw file](https://github.com/m-lab/etl/blob/acb032a27c0957032fae971fb046797729c85ab2/storage/storage.go#L44)):*

Passing the client in, or having a preallocated client in a wrapping object!  Saves a round trip to the server.  We may need to check for disconnections or some other rare conditions.
</blockquote></details>

:+1:

---

*[storage/storage.go, line 86 at r3](https://reviewable.io:443/reviews/m-lab/etl/17#-Khi-p1bRacSmFkxM9uT:-Khi-p1bRacSmFkxM9uU:b1qexi5) ([raw file](https://github.com/m-lab/etl/blob/6556af792b90cd3f95b6460866a4051efd0f1022/storage/storage.go#L86)):*
> ```Go
> 	if strings.HasSuffix(strings.ToLower(fn), "gz") {
> 		// TODO add unit test
> 		var err error
> ```

Do we need this `err`? There is one define in the outer scope, so the assignment below should be valid.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/17#-:-Khi0XiN-8WZTpB70gj6:bnfp4nl)*

gfr10598:
*[storage/storage.go, line 86 at r3](https://reviewable.io:443/reviews/m-lab/etl/17#-Khi-p1bRacSmFkxM9uT:-Khi15_c6QgQFxBNLLdj:bd30ifx) ([raw file](https://github.com/m-lab/etl/blob/6556af792b90cd3f95b6460866a4051efd0f1022/storage/storage.go#L86)):*

Do we need this `err`? There is one define in the outer scope, so the assignment below should be valid.
</blockquote></details>

good catch

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/17)*

PR #19, opened 2017-04-13
S1 ptparser

yachang:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/19)


parser/pt_parser_test.go
```diff
@@ -0,0 +1,11 @@
+package parser
+
+import (
+	"fmt"
+	"testing"
+)
+
```
gfr10598: Add TODO for IPv6 tests.

parser/pt_filename_parser.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"strings"
+)
+
+type FileName struct {
+	name string
+}
+
+// GetLocalIP parse the filename and return IP.
```
gfr10598: I think this would be much cleaner with a regexp parser.  But as this is already working, add a TODO to do that later.

parser/pt_filename_parser.go
```diff
@@ -0,0 +1,32 @@
...
+	third_segment := second_segment[secondIPStart+1 : len(second_segment)]
+	secondPortStart := strings.IndexByte(third_segment, '-')
+	secondPortEnd := strings.LastIndexByte(third_segment, '.')
+	return first_segment[0:firstPortStart], second_segment[0:secondIPStart], third_segment[0:secondPortStart], third_segment[secondPortStart+1 : secondPortEnd]
+}
+
+func (f *FileName) GetDate() string {
+	return f.name[0:8]
+}
+
```
gfr10598: Could you add a Validate method, that returns true/false, depending on whether the filename is a valid paris filename?  Suggest using a regexp.

parser/pt_parser.go
```diff
@@ -0,0 +1,69 @@
...
+type connection_spec struct {
+	server_ip      string
+	server_af      int
+	client_ip      string
+	client_af      int
+	data_direction int
+}
+
+// Save implements the ValueSaver interface.
+func (i *PT) Save() (map[string]bigquery.Value, string, error) {
```
gfr10598: FYI, though I think this should work, I ended up having to avoid pointer receiver to get a similar implementation to work with bigquery.

parser/pt_parser.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// The filename contains 5-tuple like 20170320T23:53:10Z-98.162.212.214-53849-64.86.132.75-42677.paris
+	// We can get the logtime, local IP, local port, server IP, server port from fn directly
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		oneLine := strings.TrimSuffix(scanner.Text(), "\n")
+		fmt.Println(oneLine)
+	}
+	one_row := &PT{
+		test_id:  "20170320T23:53:10Z-98.162.212.214-53849-64.86.132.75-42677.paris.gz",
```
gfr10598: If this is intended as a "trivial" parser, then name it TrivialPTParser, and make PTParser an interface?

yachang:
Review status: 0 of 6 files reviewed at latest revision, 5 unresolved discussions.

---

*[parser/pt_filename_parser.go, line 12 at r1](https://reviewable.io:443/reviews/m-lab/etl/19#-Khcm8E7ya1cWdrGs5Y7-r1-12:-KiVjteeI3KE3sINKP96:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/7241201a5286a14a0f613762e83ac4c13cc6a431/parser/pt_filename_parser.go#L12)):*

I think this would be much cleaner with a regexp parser.  But as this is already working, add a TODO to do that later.
</blockquote></details>

Done.

---

*[parser/pt_filename_parser.go, line 28 at r1](https://reviewable.io:443/reviews/m-lab/etl/19#-Khcm8E7ya1cWdrGs5Y7-r1-28:-KiVkWqbB-5GBldSfUor:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/7241201a5286a14a0f613762e83ac4c13cc6a431/parser/pt_filename_parser.go#L28)):*

Could you add a Validate method, that returns true/false, depending on whether the filename is a valid paris filename?  Suggest using a regexp.
</blockquote></details>

Done.

---

*[parser/pt_parser.go, line 32 at r1](https://reviewable.io:443/reviews/m-lab/etl/19#-Khcm8E7ya1cWdrGs5Y9-r1-32:-KiVli_WMb_sElmaUMtn:bavajrn) ([raw file](https://github.com/m-lab/etl/blob/7241201a5286a14a0f613762e83ac4c13cc6a431/parser/pt_parser.go#L32)):*

FYI, though I think this should work, I ended up having to avoid pointer receiver to get a similar implementation to work with bigquery.
</blockquote></details>

Add todo for testing BQ insertion.

---

*[parser/pt_parser.go, line 56 at r1](https://reviewable.io:443/reviews/m-lab/etl/19#-Khcm8E7ya1cWdrGs5Y9-r1-56:-KiVlu6ibc535DBLqbfg:bmornqt) ([raw file](https://github.com/m-lab/etl/blob/7241201a5286a14a0f613762e83ac4c13cc6a431/parser/pt_parser.go#L56)):*

If this is intended as a "trivial" parser, then name it TrivialPTParser, and make PTParser an interface?
</blockquote></details>

It is not a trivial parser, I will add directed tree parsing in the next PR.

---

*[parser/pt_parser_test.go, line 7 at r1](https://reviewable.io:443/reviews/m-lab/etl/19#-Khcm8E7ya1cWdrGs5YA-r1-7:-KiVm5VhxGxCW7w3xtZ9:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/7241201a5286a14a0f613762e83ac4c13cc6a431/parser/pt_parser_test.go#L7)):*

Add TODO for IPv6 tests.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/19)*

PR #23, opened 2017-04-19
Add web100 parsing library and NDTParser implementation

stephen-soltesz:
This change adds initial support for parsing NDT web100 snapshot log files from Go.

Specifically, this change includes the web100 userland source files from version 1.8 with minor fixes to support builds in 64-bit environments. To support export to BigQuery, the etl/web100 package functions return maps of the snapshot records as `map[string]bigquery.Value`. Ideally, this could be separated (pure web100-go interface and a web100-go + bigquery interface), but the current web100.c interface makes this less efficient right now.

This change includes a simple command line tool `web100_cli` that reads a web100 snapshot log file and dumps the last snapshot. This is helpful for testing.

This change includes an initial implementation of the NDTParser. It depends on the availability of a temporary directory to write snaplog files in order to use the only available parsing interface provided by the web100.c library, which operates on files.

TODO:
 * upstream the web100.c & tcp-kis.txt changes, and other inconsistencies across versions.
 * optimize web100.go implementation, e.g. to read snaplog from byte buffer instead of from file.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/23)


stephen-soltesz:
@gfr10598 I can totally split this up if it would be more helpful.

The new files and code are around 510 new lines. That is still a large change, but the bulk of this PR includes generated files or copied files from the web100 client library.

parser/ndt.go
```diff
@@ -0,0 +1,65 @@
...
+	tmpFile := fmt.Sprintf("%s/%s", n.tmpDir, fn)
+	err := ioutil.WriteFile(tmpFile, rawSnapLog, 0644)
+	if err != nil {
+		return nil, err
+	}
+	// TODO(dev): log possible remove errors.
+	defer os.Remove(tmpFile)
+
+	// Parse the tcp-kis.txt web100 variable definition file.
+	data, err := web100.Asset("tcp-kis.txt")
```
gfr10598: Add TODO to optimize this, rather than doing it every time.

stephen-soltesz: Done.

stephen-soltesz:
I've added additional TODO's around the issues I'm sure of.

I've also added a web100/README.md with steps for using `go-bindata` and a higher-level TODOs for the web100 package in general.

Regarding the current files, I think the implementation will need significant refinement but the "shape" of the web100 go interface and set of source files will probably reach production. Since web100 parsing is our core data set and the web100.c library is not ideally suited to our use case (files vs byte arrays, strings vs native types), we'll want to take additional care to guarantee that we are extracting the data we expect and that it performs comparably to the historical parser.

PR #27, opened 2017-04-21
fix out of sync parser interface in ndt

gfr10598:
Change to use Parser interface so that compiler detects interface error.
Fix interface error.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/27)

### DIFF
--- a/parser/ndt.go
+++ b/parser/ndt.go
@@ -13,10 +13,11 @@ import (
 
 type NDTParser struct {
 	Parser
-	tmpDir string
+	tmpDir    string
+	tableName string
 }
 
-func (n *NDTParser) Parse(meta map[string]bigquery.Value, testName string, table string, rawSnapLog []byte) (interface{}, error) {
+func (n *NDTParser) Parse(meta map[string]bigquery.Value, testName string, rawSnapLog []byte) (interface{}, error) {
 	// TODO(prod): do not write to a temporary file; operate on byte array directly.
 	// Write rawSnapLog to /mnt/tmpfs.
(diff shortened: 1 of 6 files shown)


parser/ndt.go
```diff
@@ -15,10 +16,10 @@ type NDTParser struct {
 	tmpDir string
 }
 
-func (n *NDTParser) Parse(fn string, table string, rawSnapLog []byte) (interface{}, error) {
+func (n *NDTParser) Parse(meta map[string]bigquery.Value, testName string, table string, rawSnapLog []byte) (interface{}, error) {
```
yachang: Add a TODO: make TableID as global const var, instead of passing it as input param everywhere.

yachang:
LGTM

gfr10598:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[parser/ndt.go, line 19 at r1](https://reviewable.io:443/reviews/m-lab/etl/27#-KiGM0EZS9iKoIvGlmNy-r1-19:-KiGYp7FxbCPX9pjnZq1:bmo3no8) ([raw file](https://github.com/m-lab/etl/blob/83fcd6876a00d6f46cf5fb2946a79f19fc86e850/parser/ndt.go#L19)):*

Add a TODO: make TableID as global const var, instead of passing it as input param everywhere.
</blockquote></details>

Added TableName to interface.  Not sure what you mean by global const var, but it shouldn't be global because it may be different for different parsers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/27)*

PR #68, opened 2017-05-10
Improve gcs retry code and add metric.

gfr10598:
This moves the GCS read retries into NextTest, with some modest changes to the retry logic.
It also adds a GCSRetryCount metric.

It retains only two log messages - one for errors creating a gzip reader (which we haven't seen yet), and one when the read fails 10 times in a row.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/68)


storage/storage.go
```diff
@@ -36,40 +38,81 @@ type ETLSource struct {
 
 // Next reads the next test object from the tar file.
 // Returns io.EOF when there are no more tests.
-func (rr *ETLSource) NextTest() (string, []byte, error) {
+func (rr *ETLSource) NextTest() (string, []byte, int, error) {
```
pboothe: At some point the size of the tuple becomes large enough that it should be a struct. Depending on your sense of taste, that point is either 3, 4, or 5 elements. I'll leave the choice up to you, but returning a 4-element tuple does not feel good. I think my preference would be returning a `(Test, error)` tuple at this point.

gfr10598:
PTAL.

---

Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[storage/storage.go, line 41 at r1](https://reviewable.io:443/reviews/m-lab/etl/68#-KjmvtqqTH3onKAAzN8G-r1-41:-KjnKEED_iQyJy9yvnwj:bfm49rb) ([raw file](https://github.com/m-lab/etl/blob/0385c48ba0c0fe77971b03cee4a99cc8cce1b722/storage/storage.go#L41)):*

At some point the size of the tuple becomes large enough that it should be a struct. Depending on your sense of taste, that point is either 3, 4, or 5 elements. I'll leave the choice up to you, but returning a 4-element tuple does not feel good. I think my preference would be returning a `(Test, error)` tuple at this point.
</blockquote></details>

Removed the new "trial" return value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/68#-:-KjnK_RHtq1hrYbnZf9B:b-83330m)*

storage/storage.go
```diff
@@ -40,36 +42,88 @@ func (rr *ETLSource) NextTest() (string, []byte, error) {
...
-	if strings.HasSuffix(strings.ToLower(h.Name), "gz") {
-		// TODO add unit test
-		zipReader, err := gzip.NewReader(rr)
+	// Try to get the next file.  We retry multiple times, because sometimes
+	// GCS stalls and produces stream errors.
+	// TODO - keep track of elapsed time instead of trials ??
+	trial := 0
+	delay := 5 * time.Millisecond
+	phase := "Next"
+	for {
```
pboothe: This loop is 80 lines and contains multiple `continue`s nested really deeply in it. I agree that it works, but would really like it to be simpler and clearer if possible.

gfr10598:
PTAL.

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[storage/storage.go, line 51 at r2](https://reviewable.io:443/reviews/m-lab/etl/68#-KjmvtqqTH3onKAAzN8G-r2-51:-KjnwloajZP6H1JtFbPq:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/57744fa5d8ab0460c0e5c94981939b63c85f6103/storage/storage.go#L51)):*

This loop is 80 lines and contains multiple `continue`s nested really deeply in it. I agree that it works, but would really like it to be simpler and clearer if possible.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/68#-:-KjnwnlRuF3pLo-wP2Td:b-83330m)*

pboothe:
Looks much better to me. Thanks! Still approved, but now with a much cleaner conscience.

LMBTM

PR #72, opened 2017-05-11
Defer processing until metafile available

gfr10598:
Add logic to cache c2s and s2c files until meta file is available, but process c2s/s2c asap once meta file is processed.

Still TODO - improve error handling with this deferred processing.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/72)


stephen-soltesz:
:lgtm:

---

Reviewed 6 of 8 files at r1.
Review status: 6 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[parser/ndt.go, line 113 at r1](https://reviewable.io:443/reviews/m-lab/etl/72#-KjtmaGrgPh4FfNzQWhA:-KjtmaGrgPh4FfNzQWhB:bym27ol) ([raw file](https://github.com/m-lab/etl/blob/74799455056ab1185b05374c92c9d5cfa7de5dd5/parser/ndt.go#L113)):*
> ```Go
> 		// a data file, or ONLY a meta file, we should log and count that.
> 		// TODO Log/count if we never see one of the three files.
> 		n.timestamp = info.Time
> ```

Just clarifying: we determined that files are in sorted order? And so are archived logs (or that we aren't concerned about those right now)?

---

*[parser/ndt.go, line 152 at r1](https://reviewable.io:443/reviews/m-lab/etl/72#-Kjtorj9HmBhrnCOQ6wj:-Kjtorj9HmBhrnCOQ6wk:b-m255l7) ([raw file](https://github.com/m-lab/etl/blob/74799455056ab1185b05374c92c9d5cfa7de5dd5/parser/ndt.go#L152)):*
> ```Go
> 			s2cErr := n.processTest(meta, n.s2c.fn, testType, n.s2c.data)
> 			if s2cErr != nil {
> 				return s2cErr
> ```

Do we care about the case when both `err` and `s2cErr` are not equal to nil?

---

*[parser/ndt_test.go, line 96 at r1](https://reviewable.io:443/reviews/m-lab/etl/72#-KjtpkrssqqBtUVuCNUH:-KjtpkrssqqBtUVuCNUI:b-iktloe) ([raw file](https://github.com/m-lab/etl/blob/74799455056ab1185b05374c92c9d5cfa7de5dd5/parser/ndt_test.go#L96)):*
> ```Go
> 			},
> 			"connection_spec": map[string]bigquery.Value{
> 				"local_port": int64(40105),
> ```

Where does this port value come from? And, why is it different from the port on the s2c_snaplog filename? Do these not actually correspond?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/72#-:-KjtqAd6QHSil4IklcXr:bnfp4nl)*

gfr10598:
Review status: 6 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[parser/ndt.go, line 113 at r1](https://reviewable.io:443/reviews/m-lab/etl/72#-KjtmaGrgPh4FfNzQWhA:-KjwtsX_mP2mlmdG87If:b-7u15cu) ([raw file](https://github.com/m-lab/etl/blob/74799455056ab1185b05374c92c9d5cfa7de5dd5/parser/ndt.go#L113)):*

Just clarifying: we determined that files are in sorted order? And so are archived logs (or that we aren't concerned about those right now)?
</blockquote></details>

I don't know about the archived logs, and that is a concern.  I'm going to add logs and metrics to detect test files without meta files, so that will also expose this, but I'll add a TODO to verify that archives are sorted, too.

---

*[parser/ndt.go, line 152 at r1](https://reviewable.io:443/reviews/m-lab/etl/72#-Kjtorj9HmBhrnCOQ6wj:-Kjwu8xTF079n3-8hp1O:b-2qitsa) ([raw file](https://github.com/m-lab/etl/blob/74799455056ab1185b05374c92c9d5cfa7de5dd5/parser/ndt.go#L152)):*

Do we care about the case when both `err` and `s2cErr` are not equal to nil?
</blockquote></details>

Yes we do, just haven't gotten to that yet.  will add TODO.

---

*[parser/ndt_test.go, line 96 at r1](https://reviewable.io:443/reviews/m-lab/etl/72#-KjtpkrssqqBtUVuCNUH:-KjwxAif3eRblsGMvfmd:b-t6mxjd) ([raw file](https://github.com/m-lab/etl/blob/74799455056ab1185b05374c92c9d5cfa7de5dd5/parser/ndt_test.go#L96)):*

Where does this port value come from? And, why is it different from the port on the s2c_snaplog filename? Do these not actually correspond?
</blockquote></details>

Heh.
Local port presumably comes from data in the file.  The thing that bothers me now that you bring my attention to it is the RemAddress,.  I didn't change that, but I see that a huge fraction of our tests are against eb.measurementlab.net and that is what I chose, and matches files you were using for testing.  

In fact, the tar file I picked only has two or three different endpoint names in it.  Huh.  Bugging me now.

Adding a TODO up above to look into this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/72)*

PR #75, opened 2017-05-17
Small tweaks

gfr10598:
Some small cleanups:
Add proper handling of .gz files (since Peter's scraper change)
Handle errors locally instead of returning to caller.
Modify backoff times for retries.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/75)

### DIFF
--- a/parser/ndt.go
+++ b/parser/ndt.go
@@ -130,7 +130,6 @@ func (n *NDTParser) reportAnomolies() {
 
 // ParseAndInsert extracts the last snaplog from the given raw snap log.
 // Writes rawSnapLog to /mnt/tmpfs.
-// TODO(dev) This is getting big and ugly and needs to be refactored.
 // TODO(prod): do not write to a temporary file; operate on byte array directly.
 func (n *NDTParser) ParseAndInsert(meta map[string]bigquery.Value, testName string, content []byte) error {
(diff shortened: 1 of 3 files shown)


stephen-soltesz:
:lgtm:

Just clarifying about the long term plan: the PR description says "Handle errors locally instead of returning to caller." I see log messages, but these are largely advisory. My understanding is that over time we'll add new handling of these events to create a persistent record of "bad files" or "bad tests".

---

Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[storage/storage.go, line 153 at r1](https://reviewable.io:443/reviews/m-lab/etl/75#-KkN0Kd1TGKHlsoHPpCK:-KkN0Kd1TGKHlsoHPpCL:b-6u33sx) ([raw file](https://github.com/m-lab/etl/blob/1eb5ddbd85ce7dd3eebfd1be73f69093ea1826a6/storage/storage.go#L153)):*
> ```Go
> 				// nextData phase of reading, but then persist on
> 				// the next call to nextHeader.
> 				break
> ```

The case where `err != nil` here would be lost. Is this intentional?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/75#-:-KkN-deFGsZNkK8bN8-1:bm1cwjs)*

PR #76, opened 2017-05-17
Use date partitions

gfr10598:
Add handling of bigquery partitions.
BQ allows streaming into partitions up to 30 days in the past, and 7 days into the future.  Outside that range, we use template table creation to write into a separate table with an _YYYYMMDD suffix.

Also updates the metrics

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/76)

### DIFF
--- a/bq/bq_test.go
+++ b/bq/bq_test.go
@@ -32,6 +32,39 @@ type Item struct {
 	Foobar int `json:"foobar"`
 }
 
+//==================================================================================
+// These tests hit the backend, to verify expected behavior of table creation and
+// access to partitions.  They deliberately have a leading "x" to prevent running
+// them in Travis.  We need to find a better way to control whether they run or
+// not.
+//==================================================================================
+func xTestRealPartitionInsert(t *testing.T) {
+	tag := "new"
+	items := []interface{}{
+		Item{Name: tag + "_x0", Count: 17, Foobar: 44},
+		Item{Name: tag + "_x1", Count: 12, Foobar: 44}}
+
+	in, err := bq.NewBQInserter(
+		etl.InserterParams{"mlab_sandbox", "test2", "_20160201", 10 * time.Second, 1}, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err = in.InsertRow(Item{Name: tag + "_x0", Count: 17, Foobar: 44}); err != nil {
+		t.Error(err)
+	}
+	if err = in.InsertRows(items); err != nil {
+		t.Error(err)
+	}
+
+	if in.Count() != 3 {
+		t.Error("Should have inserted three rows")
+	}
+	in.Flush()
+}
+
+//==================================================================================
+
 func TestBasicInsert(t *testing.T) {
 	tag := "new"
 	items := []interface{}{
@@ -39,7 +72,7 @@ func TestBasicInsert(t *testing.T) {
 		Item{Name: tag + "_x1", Count: 12, Foobar: 44}}
 
 	in, err := fake.NewFakeInserter(
-		etl.InserterParams{"mlab_sandbox", "test2", 10 * time.Second, 1})
+		etl.InserterParams{"mlab_sandbox", "test2", "", 10 * time.Second, 1})
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -65,7 +98,7 @@ func TestBufferingAndFlushing(t *testing.T) {
 	// Set up an Inserter with a fake Uploader backend for testing.
 	// Buffer 3 rows, so that we can test the buffering.
 	in, err := fake.NewFakeInserter(
-		etl.InserterParams{"mlab_sandbox", "test2", 10 * time.Second, 3})
+		etl.InserterParams{"mlab_sandbox", "test2", "", 10 * time.Second, 3})
 	if err != nil {
 		log.Printf("%v\n", err)
 		t.Fatal()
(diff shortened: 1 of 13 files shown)


stephen-soltesz:
My biggest concern is the overhead caused by using dates as metric label values. Normally we want the cardinality of label values to be low. For historical batch processing, the number of distinct values is an anti-pattern.

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[metrics/metrics.go, line 92 at r1](https://reviewable.io:443/reviews/m-lab/etl/76#-KkNU2GrtmXz7iAER2OW:-KkNU2GsDG3aIP_wgoe1:b-y0xfud) ([raw file](https://github.com/m-lab/etl/blob/ad2ab2960d97437688032494d00943d72ab9493e/metrics/metrics.go#L92)):*
> ```Go
> 		},
> 		// ndt/pt/ss, s2c/c2s/meta, ok/reject/error/
> 		[]string{"table", "suffix", "filetype", "status"},
> ```

The "suffix" label will have dates? Every distinct set of labels is a new timeseries in prometheus. Creating new timeseries is more expensive than re-using existing ones. I think I would avoid this pattern if possible.

If we are batch processing historical data (e.g. a year), we've observed that the cloud storage transfer process does not seem to guarantee order of file dates. This could make visualizing this metric more difficult or less helpful while also stressing prometheus.

I'm imagining we'll be re-processing quiet often. Is this suffix label necessary?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/76#-:-KkNon5SJv0IEV-n_8Ir:bbhkaur)*

gfr10598:
Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[metrics/metrics.go, line 92 at r1](https://reviewable.io:443/reviews/m-lab/etl/76#-KkNU2GrtmXz7iAER2OW:-KkQr7U9CgBGS76O7mx8:bz1szu1) ([raw file](https://github.com/m-lab/etl/blob/ad2ab2960d97437688032494d00943d72ab9493e/metrics/metrics.go#L92)):*

The "suffix" label will have dates? Every distinct set of labels is a new timeseries in prometheus. Creating new timeseries is more expensive than re-using existing ones. I think I would avoid this pattern if possible.

If we are batch processing historical data (e.g. a year), we've observed that the cloud storage transfer process does not seem to guarantee order of file dates. This could make visualizing this metric more difficult or less helpful while also stressing prometheus.

I'm imagining we'll be re-processing quiet often. Is this suffix label necessary?

</blockquote></details>

I believe this will have high value for at least the first few months of operation.  I'd like to be able to get some insight into which data is causing us problems.  If we get a batch of problems when we reprocess some old month, it will be helpful to get some high level idea what the distribution of the problem was.

If we determine that it is causing prometheus resource problems, we can re-evaluate.

Another perspective - this will perhaps give us some real insight into what it takes to overload prometheus.  8-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/76)*

gfr10598:
Legit concern.  Please see response to other comment.

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/76#-:-KkQriHvyFU9uI1q63hf:bgt4bu6)*

stephen-soltesz:
:lgtm: - with the understanding that we agree that using prometheus metric labels in this way is not as preferable as a persistent audit trail mechanism.

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[metrics/metrics.go, line 92 at r1](https://reviewable.io:443/reviews/m-lab/etl/76#-KkNU2GrtmXz7iAER2OW:-KkRIPzGpCA4oeR_Nj7N:b-9boyi7) ([raw file](https://github.com/m-lab/etl/blob/ad2ab2960d97437688032494d00943d72ab9493e/metrics/metrics.go#L92)):*

I believe this will have high value for at least the first few months of operation.  I'd like to be able to get some insight into which data is causing us problems.  If we get a batch of problems when we reprocess some old month, it will be helpful to get some high level idea what the distribution of the problem was.

If we determine that it is causing prometheus resource problems, we can re-evaluate.

Another perspective - this will perhaps give us some real insight into what it takes to overload prometheus.  8-)
</blockquote></details>

We talked in person and agreed that a preferable strategy for answering the question of "what data is causing problems" will be persistent audit trails.

We can use the current labels on TestCount for now, with the understanding that it should be removed once audit trails are available.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/76#-:-KkRIwsH1EARA98xJuHb:bysvokh)*

gfr10598:
Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[metrics/metrics.go, line 92 at r1](https://reviewable.io:443/reviews/m-lab/etl/76#-KkNU2GrtmXz7iAER2OW:-KkRMPG19C4MC1N_zw7t:b7i6v4m) ([raw file](https://github.com/m-lab/etl/blob/ad2ab2960d97437688032494d00943d72ab9493e/metrics/metrics.go#L92)):*

We talked in person and agreed that a preferable strategy for answering the question of "what data is causing problems" will be persistent audit trails.

We can use the current labels on TestCount for now, with the understanding that it should be removed once audit trails are available.
</blockquote></details>

Created an ETL issue to track this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/76)*

PR #116, opened 2017-05-30
Pure Go Web100 log reader

gfr10598:
Adds a pure go implementation of the web100 log file reader.
Any suggestions for a better name for the file?

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/116)


stephen-soltesz:
Thank you for taking the time to start this. A native, in-memory implementation will be a great improvement.

---

Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, 6 unresolved discussions.

---

*[web100/pure.go, line 66 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQESvE2ENAUmDYzjnI:-KlQESvE2ENAUmDYzjnJ:b-wmgvdo) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L66)):*
> ```Go
> 
> // Once consuming the header, we know the names and sizes of all fields, and the
> // size of each record, which is len(BEGIN_SNAP_DATA) + 1 + sum(field lengths)
> ```

Please use `Record` or `Snapshot` throughout. My preference is to preserve existing terminology.

IIUC, the web100.c term for a "record" is a "snapshot". "Record" is introducing new terminology. If we are creating the new reference implementation then new terminology may be fine. But, mixing terms right now is confusing.

---

*[web100/pure.go, line 66 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQIRnTUJS8mu7QRJKU:-KlQIRnTUJS8mu7QRJKV:b-e56a4z) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L66)):*
> ```Go
> 
> // Once consuming the header, we know the names and sizes of all fields, and the
> // size of each record, which is len(BEGIN_SNAP_DATA) + 1 + sum(field lengths)
> ```

Just posing a question. From early work with the web100 library we found some data was corrupted past a certain point. I'm imagining that one of the snapshots was corrupted and subsequent snapshots were all bad. How will this library contend with this?

---

*[web100/pure.go, line 84 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQV3DXFSq1F243gXVy:-KlQV3DY0o_S6l2u1HB2:b5ppkuc) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L84)):*
> ```Go
> type Header struct {
> 	Fields       []Variable
> 	RecordLength int // Total length of record, including BEGIN_SNAP_DATA
> ```

I'm not completely clear what `RecordLength` refers to. Using C types as a reference point, this is the `size_t` of `web100_snapshot.data`?

---

*[web100/pure.go, line 112 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQTxTiggx_Rau5GnYH:-KlQTxTjWVj4FErJaC26:b-lrt2nf) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L112)):*
> ```Go
> 
> // Returns the snapshot at index n, or possibly error if n is not a valid index.
> func (log *SnapLog) Snapshot(n int) (Snapshot, error) {
> ```

Hmm, I'd like a better sense of how the new interface will work in action. What do you think of adding a new file with pseudo code or example test code that uses the new interface?

For example, here I want to ask for a familiar iterator interface, but see that a direct index approach may be more helpful. In this case, getting the number of Snapshots could be helpful. e.g. maybe `log.Length()`?

---

*[web100/pure.go, line 118 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQH6K9-D_aX3ktNINx:-KlQH6K9-D_aX3ktNINy:bkashhf) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L118)):*
> ```Go
> // Convert to map suitable for writing to bigquery.
> // TODO - may drop the canonical param, and incorporate that into the Header.
> func (log *Snapshot) ToMap(canonical map[string]string) (schema.Web100ValueMap, error) {
> ```

As a low-level library meant to replace web100.c, I would like to suggest avoiding dependencies on external packages like `schema`.

---

*[web100/pure.go, line 118 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQI902XDawhyKTdJm0:-KlQI902XDawhyKTdJm1:bfynf0o) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L118)):*
> ```Go
> // Convert to map suitable for writing to bigquery.
> // TODO - may drop the canonical param, and incorporate that into the Header.
> func (log *Snapshot) ToMap(canonical map[string]string) (schema.Web100ValueMap, error) {
> ```

The receiver name should probably be `snap` instead of `log`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/116#-:-KlQYVmYtW9PULJRwws7:b-iop7bm)*

web100/pure.go
```diff
@@ -0,0 +1,469 @@
...
+	BEGIN_SNAP_DATA   = "----Begin-Snap-Data----\n"         // Plus a newline?
+	END_OF_HEADER     = "\x00----End-Of-Header---- -1 -1\n" // No newline.
+	GROUPNAME_LEN_MAX = 32
+	VARNAME_LEN_MAX   = 32
+)
+
+type VarType int
+
+const (
+	WEB100_TYPE_INTEGER VarType = iota
```
pboothe: You actually care about the values here. This isn't just an enum, it is an enum with a required order. A comment to that effect would be nice.

web100/pure.go
```diff
@@ -0,0 +1,469 @@
...
+	WEB100_ADDRTYPE_IPV6
+	WEB100_ADDRTYPE_DNS = 16
+)
+
+const (
+	WEB100_TYPE_IP_ADDRESS = WEB100_TYPE_INET_ADDRESS_IPV4 /* Deprecated */
+	WEB100_TYPE_UNSIGNED16 = WEB100_TYPE_INET_PORT_NUMBER  /* Deprecated */
+)
+
+var Web100Sizes = [WEB100_NUM_TYPES + 1]byte{
```
pboothe: A mapping type would be clearer, but I freely admit that the clarity would come with a computational hit in one of our tightest loops and so leave it up to your judgement.  Perhaps a function that is just a big switch?

web100/pure_test.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (s SimpleSaver) SetString(name string, val string) {
+	s.Strings[name] = val
+}
+
+func (s SimpleSaver) SetInt64(name string, val int64) {
+	s.Integers[name] = val
+}
+
```
pboothe: Please also include test data from 2009, to ensure that this parser also works on our oldest data.

gfr10598:
Review status: 2 of 13 files reviewed at latest revision, 9 unresolved discussions.

---

*[web100/pure.go, line 66 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQESvE2ENAUmDYzjnI:-Kl_l5qhD93LTyyPhYZ2:b-eb9dbz) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L66)):*

Please use `Record` or `Snapshot` throughout. My preference is to preserve existing terminology.

IIUC, the web100.c term for a "record" is a "snapshot". "Record" is introducing new terminology. If we are creating the new reference implementation then new terminology may be fine. But, mixing terms right now is confusing.
</blockquote></details>

I am now only using "record" as part of comments, where the context is hopefully clear.

---

*[web100/pure.go, line 66 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQIRnTUJS8mu7QRJKU:-Kl_lC7mIZzItMNNO6ZU:bj3170w) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L66)):*

Just posing a question. From early work with the web100 library we found some data was corrupted past a certain point. I'm imagining that one of the snapshots was corrupted and subsequent snapshots were all bad. How will this library contend with this?
</blockquote></details>

I've added a TODO about this.

---

*[web100/pure.go, line 84 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQV3DXFSq1F243gXVy:-Kl_lFHulveWwLr5AyYc:bzfcn9s) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L84)):*

I'm not completely clear what `RecordLength` refers to. Using C types as a reference point, this is the `size_t` of `web100_snapshot.data`?
</blockquote></details>

Changed to length and added better comment.

---

*[web100/pure.go, line 112 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQTxTiggx_Rau5GnYH:-Kl_lfX6RSLsYaX73qv7:b-ggrqfu) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L112)):*

Hmm, I'd like a better sense of how the new interface will work in action. What do you think of adding a new file with pseudo code or example test code that uses the new interface?

For example, here I want to ask for a familiar iterator interface, but see that a direct index approach may be more helpful. In this case, getting the number of Snapshots could be helpful. e.g. maybe `log.Length()`?
</blockquote></details>

Hopefully it is a bit clearer now, but let me know if you want more comments.

---

*[web100/pure.go, line 118 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQH6K9-D_aX3ktNINx:-Kl_lr29cYtAOJhmHvbv:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L118)):*

As a low-level library meant to replace web100.c, I would like to suggest avoiding dependencies on external packages like `schema`.
</blockquote></details>

Done.

---

*[web100/pure.go, line 118 at r1](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQI902XDawhyKTdJm0:-Kl_ltgHKjgV3Rg8zAbx:b-v2l2n6) ([raw file](https://github.com/m-lab/etl/blob/1853022a99d8ee3613fc4401a2dea38e5717c262/web100/pure.go#L118)):*

The receiver name should probably be `snap` instead of `log`.
</blockquote></details>

I'm reserving snap for snapshots, and using slog for SnapLog receivers.

---

*[web100/pure.go, line 59 at r2](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQ6LL5vJNbnA9Fzf6q-r2-59:-Kl_kc2AFLJxDyjQaSdh:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/a00d0cae3e74f72052405f590df2fbfb8ff0f0dd/web100/pure.go#L59)):*

You actually care about the values here. This isn't just an enum, it is an enum with a required order. A comment to that effect would be nice.
</blockquote></details>

Done.

---

*[web100/pure_test.go, line 79 at r2](https://reviewable.io:443/reviews/m-lab/etl/116#-KlRSzG8EDljfNOtTVAO-r2-79:-Kl_lzW9YukeP0y-vnOu:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/a00d0cae3e74f72052405f590df2fbfb8ff0f0dd/web100/pure_test.go#L79)):*

Please also include test data from 2009, to ensure that this parser also works on our oldest data.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/116)*

gfr10598:
Review status: 2 of 13 files reviewed at latest revision, 9 unresolved discussions.

---

*[web100/pure.go, line 89 at r2](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQ6LL5vJNbnA9Fzf6q-r2-89:-Kl_m9H6OWZeWYukwuuH:b-8b1kps) ([raw file](https://github.com/m-lab/etl/blob/a00d0cae3e74f72052405f590df2fbfb8ff0f0dd/web100/pure.go#L89)):*

A mapping type would be clearer, but I freely admit that the clarity would come with a computational hit in one of our tightest loops and so leave it up to your judgement.  Perhaps a function that is just a big switch?
</blockquote></details>

Ack.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/116)*

gfr10598:
PTAL.
ndt.go changes in a followup CL, but already working and benchmarked.

stephen-soltesz:
I'm about 50% through to the point I marked. I'll continue from there in the morning.

---

Review status: 2 of 13 files reviewed at latest revision, 15 unresolved discussions.

---

*[web100/pure.go, line 89 at r2](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQ6LL5vJNbnA9Fzf6q-r2-89:-Kl_qXf03fkEPGnJxumO:b68d1jk) ([raw file](https://github.com/m-lab/etl/blob/a00d0cae3e74f72052405f590df2fbfb8ff0f0dd/web100/pure.go#L89)):*

Ack.
</blockquote></details>

Is this var still necessary? I don't see it used elsewhere.

---

*[web100/pure.go, line 29 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kl_oMORbZZ7I018EVMA:-Kl_oMORbZZ7I018EVMB:b-et3x50) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L29)):*
> ```Go
> //   slog or snaplog - variable name for a SnapLog.
> 
> // SnapLog files look like:
> ```

Just as an aside, this text portion looks identical to the content of `/proc/web100/header`.

Based on other comments in person, it's sounding like the contents of a snaplog file are taken directly from the kernel's text or binary representations of the various structures and written directly to the snaplog only adding intermediate "End-Of-Header" and "Begin-Snap-Data" strings.

---

*[web100/pure.go, line 147 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kl_rZJ1W9DGHJFZTlM1:-Kl_rZJ1W9DGHJFZTlM2:bjxkexm) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L147)):*
> ```Go
> }
> 
> func Newvariable(s *string) (*variable, error) {
> ```

I see that `variable` is declared private with lowercase, but camel case here would still be idomatic and read more easily, I think.

Does `Newvariable` need to be public if it's returning a private type?

---

*[web100/pure.go, line 157 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kl_vdPjEU1LvASlwFgG:-Kl_vdPkJB9tKwqbTAzv:b-mhqljn) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L157)):*
> ```Go
> 		return nil, err
> 	}
> 	if varType(typ) > WEB100_TYPE_OCTET || varType(typ) < WEB100_TYPE_INTEGER {
> ```

It's possible to start at `iota - 1`. And, if we give the first and last const meaningful names then this condition could be self-describing:

```
type varType int
 
const (
	WEB100_MIN_TYPE varType = iota - 1
	WEB100_TYPE_INTEGER
       ...
        WEB100_TYPE_OCTET 
        WEB100_MAX_TYPE
)

...

    if varType(typ) >= WEB100_MAX_TYPE || varType(typ) <= WEB100_MIN_TYPE {
```

---

*[web100/pure.go, line 161 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kl_xJXhxLok7ReyyE6A:-Kl_xJXhxLok7ReyyE6B:b2ngy44) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L161)):*
> ```Go
> 	}
> 	vt := varType(typ)
> 	if length < 1 || length > 17 {
> ```

17?

---

*[web100/pure.go, line 201 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla0fAVUJtrctIUJrhw:-Kla0fAVUJtrctIUJrhx:bxja99x) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L201)):*
> canonicalName := v.Name
> 	if legacy, ok := LegacyNames[canonicalName]; ok {
> 		canonicalName = legacy
> 	}

I see this is adapted from the original logic in web100.go. My variable name choices may have been unhelpful. Just to reiterate the original intention: `LegacyNames` maps legacy names (keys) to their canonical counter part (values) e.g.
```
   legacyNames := map[string]string{
      "old": "new",
   }
```

I'm saying all this because the use of `legacy` in the new logic is a misnomer. The original logic was trying to say "assume that we've got a canonical name, unless it's present in the legacyNames map, and in that case, use the canonical name stored there."

Would it have been clearer if the map name was `CanonicalNames`?

---

*[web100/pure.go, line 208 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla5LZtVj5uiY5b4P-a:-Kla5LZuCGORGFv_C-ep:b-m3q00p) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L208)):*
> val := binary.LittleEndian.Uint32(data)
> 		if val >= 1<<31 {
> 			snapValues.SetInt64(canonicalName, int64(val)-(int64(1)<<32))
> 		} else {
> 			snapValues.SetInt64(canonicalName, int64(val))
> 		}

I cannot confirm that the additional adjustment is necessary.

What is the expected result of:

```
	data := []byte{0x00, 0x00, 0x00, 0xf0}
	val := binary.LittleEndian.Uint32(data)
	if val >= 1<<31 {
		fmt.Printf("val greater1: 0x%x\n", int64(val)-(int64(1)<<32))
		fmt.Printf("val greater2: 0x%x\n", int64(val))
	} else {
		fmt.Printf("val lesser:  0x%x\n", int64(val))
	}

...
val greater1: 0x-10000000
val greater2: 0xf0000000
```

---

*[web100/pure.go, line 212 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla7AUvweF8Qe7KoHDc:-Kla7AUvweF8Qe7KoHDd:b-co822z) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L212)):*
> ```Go
> 			snapValues.SetInt64(canonicalName, int64(val))
> 		}
> 	case WEB100_TYPE_INET_ADDRESS_IPV4:
> ```

Both type values for v6 are below. Are both types for v4 also needed? i.e. `WEB100_TYPE_IP_ADDRESS`

---

*[web100/pure.go, line 238 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla8Hju-oLkxQuCgNqB:-Kla8Hju-oLkxQuCgNqC:b8h436q) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L238)):*
> ```Go
> 		snapValues.SetString(canonicalName, ip.String())
> 	case WEB100_TYPE_STR32:
> 		// TODO - is there a better way?
> ```

How common is TYPE_STR32 in our dataset?

The SplitN is nice and very concise.

---

*[web100/pure.go, line 249 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla8nexqfAG3jPHmyhZ:-Kla8nexqfAG3jPHmyh_:b-77jfet) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L249)):*
> ```Go
> }
> 
> //=================================================================================
> ```

TODO(soltesz): continue review from here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/116#-:-Kla9D_dx0btIxLEFtNu:b-2yymi0)*

stephen-soltesz:
Review status: 2 of 13 files reviewed at latest revision, 15 unresolved discussions.

---

*[web100/pure.go, line 208 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla5LZtVj5uiY5b4P-a:-KlaRFoRuPboHHxUKyPx:b-tc78zj) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L208)):*

> val := binary.LittleEndian.Uint32(data)
> 		if val >= 1<<31 {
> 			snapValues.SetInt64(canonicalName, int64(val)-(int64(1)<<32))
> 		} else {
> 			snapValues.SetInt64(canonicalName, int64(val))
> 		}

I cannot confirm that the additional adjustment is necessary.

What is the expected result of:

```
	data := []byte{0x00, 0x00, 0x00, 0xf0}
	val := binary.LittleEndian.Uint32(data)
	if val >= 1<<31 {
		fmt.Printf("val greater1: 0x%x\n", int64(val)-(int64(1)<<32))
		fmt.Printf("val greater2: 0x%x\n", int64(val))
	} else {
		fmt.Printf("val lesser:  0x%x\n", int64(val))
	}

...
val greater1: 0x-10000000
val greater2: 0xf0000000
```
</blockquote></details>

D'oh. Nevermind this. I was focusing on the code inside the case and overlooked the type name `WEB100_TYPE_INTEGER32`. I got focused on the `Uint32` as though that was the type.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/116)*

gfr10598:
Review status: 2 of 13 files reviewed at latest revision, 15 unresolved discussions.

---

*[web100/pure.go, line 147 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kl_rZJ1W9DGHJFZTlM1:-KlaEmiIRB3eV_xyUM1p:blu6bld) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L147)):*

I see that `variable` is declared private with lowercase, but camel case here would still be idomatic and read more easily, I think.

Does `Newvariable` need to be public if it's returning a private type?
</blockquote></details>

Oops - that was a global find/replace.  Fixed.

---

*[web100/pure.go, line 161 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kl_xJXhxLok7ReyyE6A:-KlaGbOBMvYoC7F8BsKP:b4sv255) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L161)):*

17?
</blockquote></details>

Fixed.  Thanks!

---

*[web100/pure.go, line 201 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla0fAVUJtrctIUJrhw:-KlaHCOXmGW0zHa0rVVf:b97fp5m) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L201)):*

> canonicalName := v.Name
> 	if legacy, ok := LegacyNames[canonicalName]; ok {
> 		canonicalName = legacy
> 	}

I see this is adapted from the original logic in web100.go. My variable name choices may have been unhelpful. Just to reiterate the original intention: `LegacyNames` maps legacy names (keys) to their canonical counter part (values) e.g.
```
   legacyNames := map[string]string{
      "old": "new",
   }
```

I'm saying all this because the use of `legacy` in the new logic is a misnomer. The original logic was trying to say "assume that we've got a canonical name, unless it's present in the legacyNames map, and in that case, use the canonical name stored there."

Would it have been clearer if the map name was `CanonicalNames`?
</blockquote></details>

Changed to CanonicalNames.  Good idea.

---

*[web100/pure.go, line 208 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla5LZtVj5uiY5b4P-a:-KlaQxbdhI8QVv8unXCc:b-vdhguk) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L208)):*

> val := binary.LittleEndian.Uint32(data)
> 		if val >= 1<<31 {
> 			snapValues.SetInt64(canonicalName, int64(val)-(int64(1)<<32))
> 		} else {
> 			snapValues.SetInt64(canonicalName, int64(val))
> 		}

I cannot confirm that the additional adjustment is necessary.

What is the expected result of:

```
	data := []byte{0x00, 0x00, 0x00, 0xf0}
	val := binary.LittleEndian.Uint32(data)
	if val >= 1<<31 {
		fmt.Printf("val greater1: 0x%x\n", int64(val)-(int64(1)<<32))
		fmt.Printf("val greater2: 0x%x\n", int64(val))
	} else {
		fmt.Printf("val lesser:  0x%x\n", int64(val))
	}

...
val greater1: 0x-10000000
val greater2: 0xf0000000
```
</blockquote></details>

I'm not sure what to make of this.  I've added a unit test, and if I remove the adjustment, it fails.
I've updated to use hex values instead of 1<<31, etc.

---

*[web100/pure.go, line 212 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla7AUvweF8Qe7KoHDc:-KlaRdvsSSStgo7NIHA4:b-75r62n) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L212)):*

Both type values for v6 are below. Are both types for v4 also needed? i.e. `WEB100_TYPE_IP_ADDRESS`
</blockquote></details>

These are not equivalent, and I'm pretty sure the answer is no.  I'm removing the aliases, as they are unused, and probably just confuse things.

---

*[web100/pure.go, line 238 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla8Hju-oLkxQuCgNqB:-KlaS7Cl764en5lMOIQF:boti8ka) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L238)):*

How common is TYPE_STR32 in our dataset?

The SplitN is nice and very concise.
</blockquote></details>

Pretty sure it never occurs.  Added a note at top of function.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/116)*

gfr10598:
Review status: 2 of 13 files reviewed at latest revision, 15 unresolved discussions.

---

*[web100/pure.go, line 89 at r2](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQ6LL5vJNbnA9Fzf6q-r2-89:-KlaSv5CVWJXHgmc1hap:bojorot) ([raw file](https://github.com/m-lab/etl/blob/a00d0cae3e74f72052405f590df2fbfb8ff0f0dd/web100/pure.go#L89)):*

Is this var still necessary? I don't see it used elsewhere.
</blockquote></details>

I just added field length verification, so now it is used.

---

*[web100/pure.go, line 147 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kl_rZJ1W9DGHJFZTlM1:-KlaSpvRdoDhdoAFm0ty:b6fldro) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L147)):*

Oops - that was a global find/replace.  Fixed.
</blockquote></details>

Actually, I'm leaving it exported, for testing purposes.  See new unit test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/116)*

stephen-soltesz:
Review status: 2 of 13 files reviewed at latest revision, 13 unresolved discussions.

---

*[web100/pure.go, line 89 at r2](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQ6LL5vJNbnA9Fzf6q-r2-89:-KldTQMhw4BO-9OvRi6C:b3k0158) ([raw file](https://github.com/m-lab/etl/blob/a00d0cae3e74f72052405f590df2fbfb8ff0f0dd/web100/pure.go#L89)):*

I just added field length verification, so now it is used.
</blockquote></details>

For readability, would you be willing to put each element on a separate line?

---

*[web100/pure.go, line 208 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla5LZtVj5uiY5b4P-a:-KldQseYKC5mJKLuam-w:b-l83xo9) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L208)):*

I'm not sure what to make of this.  I've added a unit test, and if I remove the adjustment, it fails.
I've updated to use hex values instead of 1<<31, etc.

</blockquote></details>

Looks like our comments crossed paths. Just my confusion.

---

*[web100/pure.go, line 265 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldGVwK5TPDjlwprblV:-KldGVwK5TPDjlwprblW:bz03rvk) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure.go#L265)):*
> ```Go
> 
> //=================================================================================
> // ConnectionSpec is used to
> ```

Comment is incomplete.

---

*[web100/pure.go, line 327 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldJSUODbqu9HpP7HAM:-KldJSUODbqu9HpP7HAN:bc8br4) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure.go#L327)):*
> ```Go
> 	if pre != preamble {
> 		return fields, errors.New("Expected preamble: " +
> 			preamble[:len(preamble)-2] + " != " + pre[:len(pre)-2])
> ```

-2? Does this presume some knowledge of the format of preamble?

---

*[web100/pure.go, line 337 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldKJFj9rqpC5W5nSDm:-KldKJFki9YPR2FsiWNF:br05tdg) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure.go#L337)):*
> ```Go
> 				return fields, errors.New("Encountered EOF")
> 			} else {
> 				return fields, errors.New("Corrupted header")
> ```

Is performance better when returning objects rather than pointers? I'm asking b/c I'm used to the pattern of returning nil when the error is non-nil. And, this pattern raises questions about whether fields is in a consistent or inconsistent state as far as the caller is concerned. Are any errors unrecoverable? Or, should any error be treated differently by the caller than other errors? A docstring comment may be helpful.

---

*[web100/pure.go, line 368 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldLnQ38HGepwQKY9b5:-KldLnQ38HGepwQKY9b6:b-y3l09h) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure.go#L368)):*
> ```Go
> 	// WARNING - the web100 code seemingly depends on a 32 bit architecture.
> 	// There is no "packed" directive for the web100_connection_spec, and the
> 	// fields all seem to be 32 bit aligned.
> ```

Wow. Good warning.

I'm not sure of the performance cost for leveraging only C-go structs. If it were negligible, this might be a safer and/or clearer use-case by defining the C struct and reading values directly rather than by byte offsets.

---

*[web100/pure.go, line 515 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldRYa8DpD97eTDHE58:-KldRYa8DpD97eTDHE59:b-35uzlo) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure.go#L515)):*
> ```Go
> 		// Extract the web100 variable name and type. This will
> 		// correspond to one of the variables defined in tcp-kis.txt.
> 		// TODO handle canonical names
> ```

Canonical names are handled now in `Save`?

---

*[web100/pure_test.go, line 19 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldSZoaMPb0314PrmwQ:-KldSZobhjnQYHXkfqN5:bf68dkj) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure_test.go#L19)):*
> ```Go
> 
> func TestHeaderParsing(t *testing.T) {
> 	c2sName := `20170509T13:45:13.590210000Z_eb.measurementlab.net:48716.c2s_snaplog`
> ```

During our original testing we found a number of files that we believed were actually corrupted in some way. Would it be worthwhile to include a typical "corrupt" file to verify the parser ignores appropriately?

I have a longer list, and this may include some false positives, but this should be a start. prg01 was the site that had chronic corrupt files.

gs://m-lab/ndt/2016/04/01/20160401T000000Z-mlab1-prg01-ndt-0000.tgz
gs://m-lab/ndt/2016/04/01/20160401T000000Z-mlab1-prg01-ndt-0001.tgz
gs://m-lab/ndt/2016/04/01/20160401T000000Z-mlab1-prg01-ndt-0004.tgz
gs://m-lab/ndt/2016/04/01/20160401T000000Z-mlab1-prg01-ndt-0005.tgz
gs://m-lab/ndt/2016/11/01/20161101T000000Z-mlab1-syd01-ndt-0002.tgz

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/116)*

gfr10598:
PTAL

---

Review status: 2 of 13 files reviewed at latest revision, 13 unresolved discussions.

---

*[web100/pure.go, line 89 at r2](https://reviewable.io:443/reviews/m-lab/etl/116#-KlQ6LL5vJNbnA9Fzf6q-r2-89:-KlddJ9irXeHvC75cROz:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/a00d0cae3e74f72052405f590df2fbfb8ff0f0dd/web100/pure.go#L89)):*

For readability, would you be willing to put each element on a separate line?
</blockquote></details>

Done.

---

*[web100/pure.go, line 29 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kl_oMORbZZ7I018EVMA:-KleC5kupyyGliKT6c6w:b-fcgkqt) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L29)):*

Just as an aside, this text portion looks identical to the content of `/proc/web100/header`.

Based on other comments in person, it's sounding like the contents of a snaplog file are taken directly from the kernel's text or binary representations of the various structures and written directly to the snaplog only adding intermediate "End-Of-Header" and "Begin-Snap-Data" strings.
</blockquote></details>

I've add the first line as a comment.  I'm leaving the rest as is for now.

---

*[web100/pure.go, line 157 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kl_vdPjEU1LvASlwFgG:-KlddTW92utztCD3aCNs:bbz955z) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L157)):*

It's possible to start at `iota - 1`. And, if we give the first and last const meaningful names then this condition could be self-describing:

```
type varType int
 
const (
	WEB100_MIN_TYPE varType = iota - 1
	WEB100_TYPE_INTEGER
       ...
        WEB100_TYPE_OCTET 
        WEB100_MAX_TYPE
)

...

    if varType(typ) >= WEB100_MAX_TYPE || varType(typ) <= WEB100_MIN_TYPE {
```
</blockquote></details>

I guess I don't fully understand this, and it isn't clear what it's utility is.

---

*[web100/pure.go, line 249 at r3](https://reviewable.io:443/reviews/m-lab/etl/116#-Kla8nexqfAG3jPHmyhZ:-KleFeJz0xC8QACHFcgo:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/cde3f750524919b7b439b3c8ca9315e05894b991/web100/pure.go#L249)):*

TODO(soltesz): continue review from here.
</blockquote></details>

Done.

---

*[web100/pure.go, line 265 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldGVwK5TPDjlwprblV:-KleCaImSo_Pp2IO_foS:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure.go#L265)):*

Comment is incomplete.
</blockquote></details>

Done.

---

*[web100/pure.go, line 327 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldJSUODbqu9HpP7HAM:-KleDEI3WNWiWgT9HTsc:b-j99vj0) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure.go#L327)):*

-2? Does this presume some knowledge of the format of preamble?
</blockquote></details>

Should be -1, and added a comment.

---

*[web100/pure.go, line 337 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldKJFj9rqpC5W5nSDm:-KleEo5e20xxyQy0DaB5:b7a3u1q) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure.go#L337)):*

Is performance better when returning objects rather than pointers? I'm asking b/c I'm used to the pattern of returning nil when the error is non-nil. And, this pattern raises questions about whether fields is in a consistent or inconsistent state as far as the caller is concerned. Are any errors unrecoverable? Or, should any error be treated differently by the caller than other errors? A docstring comment may be helpful.
</blockquote></details>

Good point.  Probably not, and not relevant anyway.  Fixed.

---

*[web100/pure.go, line 368 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldLnQ38HGepwQKY9b5:-KleF-22ilF4-H4WY3b0:b-rtv9is) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure.go#L368)):*

Wow. Good warning.

I'm not sure of the performance cost for leveraging only C-go structs. If it were negligible, this might be a safer and/or clearer use-case by defining the C struct and reading values directly rather than by byte offsets.
</blockquote></details>

I don't think there is any advantage to using cgo here.  Prefer to make it explicit.  If there is a standard way to unmarshal here, that might be better.

---

*[web100/pure.go, line 515 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldRYa8DpD97eTDHE58:-KldeHNC5VFqMoF5FRqC:bq5wuc9) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure.go#L515)):*

Canonical names are handled now in `Save`?
</blockquote></details>

Updated comment.

---

*[web100/pure_test.go, line 19 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldSZoaMPb0314PrmwQ:-Kle8W7nVVyEZthCsruP:b-5amgbo) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure_test.go#L19)):*
> gs://m-lab/ndt/2016/04/01/20160401T000000Z-mlab1-prg01-ndt-0004.tgz
I ran the first and third through the old and new parsers, and didn't see anything noteworthy in either parser, aside from one or two oversized tests.
The syd01 file has small snaplogs, and one w.Next error (on the old parser).  The new parser correctly handles it, too.
Think it is worth the hassle to extract this test and make it a unit test?  Adding a TODO at the head of the file.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/116#-:-KleFo4lTbK2KEILLjzI:bb71pro)*

stephen-soltesz:
:lgtm:

---

Review status: 2 of 13 files reviewed at latest revision, 5 unresolved discussions.

---

*[web100/pure_test.go, line 19 at r4](https://reviewable.io:443/reviews/m-lab/etl/116#-KldSZoaMPb0314PrmwQ:-KleHTZqOfMIu0FDuXLf:b-bsrjhg) ([raw file](https://github.com/m-lab/etl/blob/03a0cc8a4de9591b2fce6d2a70188839303b5d99/web100/pure_test.go#L19)):*

> gs://m-lab/ndt/2016/04/01/20160401T000000Z-mlab1-prg01-ndt-0004.tgz
I ran the first and third through the old and new parsers, and didn't see anything noteworthy in either parser, aside from one or two oversized tests.
The syd01 file has small snaplogs, and one w.Next error (on the old parser).  The new parser correctly handles it, too.
Think it is worth the hassle to extract this test and make it a unit test?  Adding a TODO at the head of the file.

</blockquote></details>

I'll leave it to you to weigh the cost benefit. My thought was that the current tests don't cover many error paths, and having examples of typically malformed data might be helpful demonstration of the code's resilience. Not a blocker.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/116#-:-KleHwjjj0q08Ls6buV8:bnfp4nl)*

PR #127, opened 2017-06-12
new group deferal strategy and collision handling

gfr10598:
Previously, c2s and s2c tests were processed and forwarded to the inserter as soon as a meta file was available.  If a new test came in before .meta file, the tests were processed without the meta file.

This lead to orphaned tests if the tests at the end of a tar file had no meta file.

This restructures the handling of test groups, and allows explicitly processing the last group at the end of the tar file, in the Flush() function.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/127)

### DIFF
--- a/parser/ndt.go
+++ b/parser/ndt.go
@@ -155,6 +155,16 @@ func (n *NDTParser) ParseAndInsert(taskInfo map[string]bigquery.Value, testName 
 		// Handle previous test group before processing new group.
 		n.processGroup()
 
+		// Verify that tests are arriving in timestamp order.
+		// TODO(prod) Consider moving this up to task.go (or storage.go)
+		if info.Time < n.timestamp {
+			metrics.ErrorCount.WithLabelValues(
+				n.TableName(), "unknown", "TIMESTAMPS OUT OF ORDER").Inc()
+			log.Printf("Timestamps out of order in: %s\n",
+				n.taskFileName, err)
+			panic("Timestamps out of order in tar file")
+		}
+
 		n.taskFileName = taskInfo["filename"].(string)
 		n.timestamp = info.Time
 	} else {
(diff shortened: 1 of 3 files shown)


parser/ndt.go
```diff
@@ -149,22 +151,24 @@ func (n *NDTParser) ParseAndInsert(taskInfo map[string]bigquery.Value, testName
...
+		if n.taskFileName != taskInfo["filename"].(string) {
+			metrics.TestCount.WithLabelValues(
+				n.TableName(), "any", "inconsistent taskFileName").Inc()
+		}
 	}
 
+	// Because of port number, the c2s, s2c, and meta files may come in
+	// any order.  We defer processing until Flush or new test group.
+	// TODO - should we just ignore non-gzipped test files?  Or do some archives
+	// have unzipped files?
```
pboothe: It looks like gzip compression capabilities did not arrive until after MLab started ( https://github.com/ndt-project/ndt/blob/master/src/logging.c#L45 ). So I can confidently say that ignoring all non-zipped results is the wrong thing.  Ignoring unzipped files if the zipped ones are present is definitely the right thing.

parser/ndt.go
```diff
@@ -177,11 +181,10 @@ func (n *NDTParser) ParseAndInsert(taskInfo map[string]bigquery.Value, testName
 					n.TableName(), "c2s", "timestamp collision").Inc()
 				log.Printf("Collision: %s and %s\n", n.c2s.fn, testName)
 			}
-		} else {
-			// We only take action on the first file, to avoid dups.
-			n.c2s = &fileInfoAndData{testName, *info, content}
-			n.processTest(taskFileName, n.c2s, "c2s")
 		}
+		// We always use the latest file, since .gz is more reliably
+		// complete, and lexicographically later.
```
pboothe: Okay, but then this ups the priority on `TODO(prod): make sure historical tar files are also in lexicographic order`

gfr10598:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[parser/ndt.go, line 171 at r1](https://reviewable.io:443/reviews/m-lab/etl/127#-KmRs5UmJWIwaRD-vBp2-r1-171:-KmSDOtOkFTSqNf2x4oR:bxgmu3e) ([raw file](https://github.com/m-lab/etl/blob/9ac7ed9b5186df024e75f92a33617e32fd1d1a1c/parser/ndt.go#L171)):*

It looks like gzip compression capabilities did not arrive until after MLab started ( https://github.com/ndt-project/ndt/blob/master/src/logging.c#L45 ). So I can confidently say that ignoring all non-zipped results is the wrong thing.  Ignoring unzipped files if the zipped ones are present is definitely the right thing.
</blockquote></details>

Updated.  Also tweaked code below.

---

*[parser/ndt.go, line 186 at r1](https://reviewable.io:443/reviews/m-lab/etl/127#-KmRs5UmJWIwaRD-vBp2-r1-186:-KmSDVppaSruK5UgCnJY:bjdf5b7) ([raw file](https://github.com/m-lab/etl/blob/9ac7ed9b5186df024e75f92a33617e32fd1d1a1c/parser/ndt.go#L186)):*

Okay, but then this ups the priority on `TODO(prod): make sure historical tar files are also in lexicographic order`
</blockquote></details>

Updated to actually prefer the .gz explicitly.
Added check to ensure we handle non-lexical ordering cases, and to detect and report timestamps out of order.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/127)*

gfr10598:
PTAL

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/127#-:-KmSFpoPSe2SFk_HgscC:bb71pro)*

pboothe:
LGTM

PR #148, opened 2017-06-19
Count the files collected daily per host and rsync module

stephen-soltesz:
This PR adds minimal support for reconciling the number of files uploaded per day by scraper with the number of files processed per day by the etl workers. This PR is a companion to https://github.com/m-lab/scraper/pull/72

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/148)

### DIFF
--- a/cmd/etl_worker/etl_worker.go
+++ b/cmd/etl_worker/etl_worker.go
@@ -197,7 +197,12 @@ func worker(w http.ResponseWriter, r *http.Request) {
 	p := parser.NewParser(dataType, ins)
 	tsk := task.NewTask(fn, tr, p)
 
-	err = tsk.ProcessAllTests()
+	files, err := tsk.ProcessAllTests()
+
+	// Count the files processed per-host-module per-weekday.
+	metrics.FileCount.WithLabelValues(
+		data.Host+"-"+data.Pod+"-"+data.Experiment,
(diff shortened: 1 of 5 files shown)


cmd/etl_worker/etl_worker.go
```diff
@@ -197,7 +197,12 @@ func worker(w http.ResponseWriter, r *http.Request) {
 	p := parser.NewParser(dataType, ins)
 	tsk := task.NewTask(fn, tr, p)
 
-	err = tsk.ProcessAllTests()
+	files, err := tsk.ProcessAllTests()
+
+	// Count the files processed per-host-module per-weekday.
+	metrics.FileCount.WithLabelValues(
+		data.Host+"-"+data.Pod+"-"+data.Experiment,
```
gfr10598: I'd suggest having host and pod be separate labels.  It shouldn't impact the metric footprint very much, and would allow us to aggregate easily across hosts.

Up to you.  Also fine to leave a TODO.

stephen-soltesz: Good idea. I'd like to see how the current version works in practice. I'm still skeptical about the entire approach of joining daily metrics with prometheus. If the single-metric approach is indeed helpful then I'll look at separate labels here to gain the additional flexibility. Added a TODO.

Issue #168, opened 2017-06-29
Mismatches between new and old pipeline output


gfr10598:
For Nov 1, 2016, based on visual inspection of merged tables.

New parser is almost always missing:
  client_af (always 0, instead of 2)
  client_kernel_version
  client_version
  leading ndt.iupui from server_hostname
  when no_meta, the server_hostname is missing entirely.
  
New parser does not strip .gz suffix from file names.

New parser is including client_browser "- (web100clt)", but old parser filters it out.

0. log_time sometimes differs by 1 second.  Rounding??

Many fields are null in old tables, but empty string in new tables.

gfr10598:
Oops - the filenames missing .gz are from the new parser, because the tar file contained unzipped files.  The legacy table has .gz suffixes.

gfr10598:
server_hostname is still empty when meta is missing.

oversize (>10M) files actually show up in legacy, so perhaps we should parse and label them.

gfr10598:
Still seeing
* leading ndt.iupui from server_hostname (will not fix?)
* when no_meta, the server_hostname is missing entirely  (fix isn't working **********)
* .gz suffix on legacy but not new. (caused by legacy archive concentrator - will not fix)
* New parser is including client_browser "- (web100clt)", but old parser filters it out.  (will not fix)
* log_time sometimes differs by 1 second.  (will not fix)

gfr10598:
server_hostname is fixed #185

gfr10598:
>10M filed as new issue - #191

PR #172, opened 2017-06-29
Add missing NDT client values & Add tls & websocket columns

stephen-soltesz:
This PR addresses two issues:

 * client_kernel_version & client_version are parsed from .meta files addressing this issue from https://github.com/m-lab/etl/issues/168
 * "tls" and "websocket" are added as new columns with unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/172)

### DIFF
--- a/parser/ndt_meta.go
+++ b/parser/ndt_meta.go
@@ -51,8 +51,8 @@ var fieldPairs = map[string]string{
 
 	// NDT SSL added two additional meta fields to signify whether the test was
 	// a websocket and/or tls test.
-	"tls":       "tls",
-	"websocket": "websocket",
+	"tls":        "tls",
+	"websockets": "websockets",
 }
 
 func handleIP(connSpec schema.Web100ValueMap, prefix string, ipString string) {
@@ -83,8 +83,17 @@ func (mfd *MetaFileData) PopulateConnSpec(connSpec schema.Web100ValueMap) {
 			log.Printf("Missing field: %s %v\n", k, v)
 		}
 	}
-	connSpec.SetBool("tls", mfd.Tls)
-	connSpec.SetBool("websocket", mfd.Websockets)
+	// Only set the value for tls & websocket if the field is present.
+	if s, ok := mfd.Fields["tls"]; ok {
+		if s != "" {
+			connSpec.SetBool("tls", mfd.Tls)
+		}
+	}
+	if s, ok := mfd.Fields["websockets"]; ok {
+		if s != "" {
+			connSpec.SetBool("websockets", mfd.Websockets)
+		}
+	}
 	s, ok := mfd.Fields["server_ip"]
 	// TODO - extract function for this stanza
 	if ok {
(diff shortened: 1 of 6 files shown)


gfr10598:
:lgtm:
Please rebase before submitting.

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[parser/ndt_meta.go, line 80 at r1](https://reviewable.io:443/reviews/m-lab/etl/172#-KnplfFF1rjbELyCzkpl:-KnplfFF1rjbELyCzkpm:b8zq65i) ([raw file](https://github.com/m-lab/etl/blob/1848a7d8b952a6d1817713a3d2176791871a3467/parser/ndt_meta.go#L80)):*
> ```Go
> 		}
> 	}
> 	connSpec.SetBool("tls", mfd.Tls)
> ```

Should these be tri-valued, and handle the case where the meta file doesn't set them?  IIUC this code assumes they are set, so they will only be null if the meta file is missing.

---

*[parser/ndt_meta_test.go, line 47 at r1](https://reviewable.io:443/reviews/m-lab/etl/172#-Knpm7Blmkn22IEDDkeb:-Knpm7Blmkn22IEDDkec:bhm2d5y) ([raw file](https://github.com/m-lab/etl/blob/1848a7d8b952a6d1817713a3d2176791871a3467/parser/ndt_meta_test.go#L47)):*
> ```Go
> 	}
> 	if connSpec["websocket"] != true {
> 		t.Error("Incorrect websocket: got false; want")
> ```

want true?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/172#-:-KnpmKHGlKFymPETEm_0:bkvyjac)*

stephen-soltesz:
I fixed a typo for "websocket" -> "websockets" since that is the actual field name from ndt.

I believe I've preserved the tri-value state for the tls and websockets fields.

I updated `createMetaFileData` to additionally save the tls and websockets fields to `mfd.Fields`, since they were originally not saved only parsed, and for the tri-value state we need both.

Note: I have not updated the public tables with the additional columns.

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[parser/ndt_meta.go, line 80 at r1](https://reviewable.io:443/reviews/m-lab/etl/172#-KnplfFF1rjbELyCzkpl:-KnppA9zT6vLneNrzC5A:b-9u0mup) ([raw file](https://github.com/m-lab/etl/blob/1848a7d8b952a6d1817713a3d2176791871a3467/parser/ndt_meta.go#L80)):*

Should these be tri-valued, and handle the case where the meta file doesn't set them?  IIUC this code assumes they are set, so they will only be null if the meta file is missing.
</blockquote></details>

Good point. I've fixed this by making the call to `SetBool` conditional on whether the field is present and non-empty.

---

*[parser/ndt_meta_test.go, line 47 at r1](https://reviewable.io:443/reviews/m-lab/etl/172#-Knpm7Blmkn22IEDDkeb:-KnppGWCNFHQEa-1LibZ:bt2vr1b) ([raw file](https://github.com/m-lab/etl/blob/1848a7d8b952a6d1817713a3d2176791871a3467/parser/ndt_meta_test.go#L47)):*

want true?
</blockquote></details>

Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/172#-:-KnpyTxf8r8spl7e8uHt:bos4byw)*

gfr10598:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/172#-:-Knq0pP3Ip6NX78vLJF-:bnfp4nl)*

PR #185, opened 2017-06-30
Fix check for server_hostname

stephen-soltesz:
Since the connSpec is a `map[string]bigquery.Value` the empty value is `nil` if `server_hostname` is not present. So this change explicitly checks whether `server_hostname` is present and if it is that it is not empty. If it is missing or empty, the value is constructed and set.

Addresses issue from: https://github.com/m-lab/etl/issues/168

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/185)

### DIFF
--- a/parser/ndt.go
+++ b/parser/ndt.go
@@ -562,13 +562,17 @@ func (n *NDTParser) fixValues(r schema.Web100ValueMap) {
 	}
 
 	// If there is no meta file then the server hostname will not be set.
-	if connSpec["server_hostname"] == "" {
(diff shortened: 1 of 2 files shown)


gfr10598:
:lgtm:

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/185#-:-Knthdl32Mov5-OGHFFf:bnfp4nl)*

PR #215, opened 2017-08-01
Sidestream perf tuning

yachang:
After tuning, CPU utilization > 90%
tooManyRequest error drop from thousands to single digit
50% insertion take about 1s.

### DIFF
--- a/cmd/etl_worker/app-sidestream.yaml
+++ b/cmd/etl_worker/app-sidestream.yaml
@@ -15,18 +15,10 @@ resources:
   # TODO - Adjust once we understand requirements.
   disk_size_gb: 10
 
-  # The tmpfs space is allocated from memory_gb.
-  volumes:
-  # Define a tmpfs disk for NDT snapshot files.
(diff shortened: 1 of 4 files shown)


cmd/etl_worker/app-sidestream.yaml
```diff
@@ -15,18 +15,10 @@ resources:
...
-  # Mounted under: /mnt/tmpfs
-  - name: tmpfs
-    volume_type: tmpfs
-    size_gb: 1.0
-
 automatic_scaling:
   # We expect fairly steady load, so a modest minimum will rarely cost us anything.
   min_num_instances: 2
-  max_num_instances: 20
+  max_num_instances: 50
```
stephen-soltesz: Oh, wow. I thought the sidestream processing was fairly lightweight. Can you add a comment about why sidestream needs 50 instances?

yachang: Change it back to 20. In running, there are only 4 instances triggered even with 95% CPU utilization.

PR #252, opened 2017-09-29
Last of Joseph's changes, adding flag protection.

gfr10598:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/252)

### DIFF
--- a/parser/metadata.go
+++ b/parser/metadata.go
@@ -21,6 +21,27 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+var ipAnnotationEnabled = false
+
+func init() {
+	checkFlags()
+}
+
+func checkFlags() {
+	// Check for ANNOTATE_IP = 'true'
+	flag, ok := os.LookupEnv("ANNOTATE_IP")
+	if ok {
+		ipAnnotationEnabled, _ = strconv.ParseBool(flag)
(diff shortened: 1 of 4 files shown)


parser/metadata_test.go
```diff
@@ -376,51 +376,109 @@ func TestGetAndInsertGeolocationIPStruct(t *testing.T) {
...
-				geos["postal_code"] = "10584"
-				geos["latitude"] = float64(41.0051)
-				geos["longitude"] = float64(73.7846)
-				return spec
-			}(),
-		},
+	return tst
+}
+
+var tests = []struct {
```
pboothe: Is this an idiomatic Go thing that I'll have to get used to, or should I regard it as worryingly janky?

parser/metadata.go
```diff
@@ -21,6 +21,27 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+var ipAnnotationEnabled = false
+
+func init() {
```
pboothe: TIL about `init()` in Go.

parser/metadata_test.go
gfr10598: It is idiomatic go testing structure.
Unless you mean the anonymous struct, in which case, it is general idiomatic go.

Issue #262, opened 2017-10-09
Handle high rate insert errors


gfr10598:
If there is more than one or two insert errors for a file we should probably classify that as a task error, and fail the task.

This came up with the sandbox pipeline, when the table was accidentally deleted, but the pipeline kept working, with error on every single insert attempt.

gfr10598:
PR #274

PR #274, opened 2017-10-16
Fail tasks with high row errors

gfr10598:
Issue #262

There are two cases where we see high row insert errors, but files are not reprocessed.
1.  table is missing and all inserts fail.
2.  insert quota is exceeded, and many inserts fail.

Adds a parser.TaskError check at end of task processing.  
Adds NDT TaskError implementation that fails if more than 10% of row adds failed.

This may result in more dups, if there are spurious errors in excess of 10%, e.g. when there are lots of quota errors.

This enables, but does not address, handling of similar problems in other parsers.  Will add issues to address those.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/274)


stephen-soltesz:
Please clarify the second sentence after the list in the pr description - it was hard to parse.

From in person discussion, it sounds like there are two kinds of potential BQ insert quota errors, per-table and per-project. My concern that won't be addressed in this PR is what happens to parsers without the recovery logic when per-project quotas are reached?

:lgtm:

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/274#-:-Kw_majSwRWdmjau0-sM:bnfxwbb)*

gfr10598:
Reviewed 7 of 8 files at r1, 1 of 1 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/274)*

PR #290, opened 2017-10-24
Allow leading ./ in filenames

gfr10598:
Continuous scraper now has leading ./ in filenames.  We will want to fix this,  but allowing it now for testing.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/290)


gfr10598:
The ./ only appeared in staging, so this PR is not required for production or reprocessing.
Closing.

PR #302, opened 2017-11-14
Suppress delta generation

gfr10598:
For near term, we can produce tables without deltas about 5 times faster than with deltas.  We will provide some test tables with deltas, and add deltas back into the main tables in a couple months.

This does some refactoring, and adds an env var check that disables delta generation.  It also sets the env var for sandbox.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/302)

### DIFF
--- a/cmd/etl_worker/app-ndt.yaml
+++ b/cmd/etl_worker/app-ndt.yaml
@@ -41,4 +41,5 @@ env_variables:
   BIGQUERY_PROJECT: 'mlab-sandbox'
   BIGQUERY_DATASET: 'mlab_sandbox'
   ANNOTATE_IP: 'true'
+  NDT_OMIT_DELTAS: 'true'
   # TODO add custom service-account, instead of using default credentials.
--- a/parser/ndt.go
+++ b/parser/ndt.go
@@ -20,7 +20,7 @@ import (
 )
 
 var (
-	NDTOmitDeltas = os.GetEnv("NDT_OMIT_DELTAS")
+	NDTOmitDeltas = os.Getenv("NDT_OMIT_DELTAS")
 )
 
 const (
--- a/parser/ndt.go
+++ b/parser/ndt.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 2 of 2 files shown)


pboothe:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[parser/ndt.go, line 25 at r3](https://reviewable.io:443/reviews/m-lab/etl/302#-KyvJAGlO24Hqn0SnXU5:-KyvJAGmP8QonOHcM7N3:b-88q5wl) ([raw file](https://github.com/m-lab/etl/blob/ce13e46d7bf777cac03f93c942c068fb3ba7400c/parser/ndt.go#L25)):*
> ```Go
> var (
> 	// NDTOmitDeltas flag indicates if deltas should be suppressed.
> 	NDTOmitDeltas, _ = strconv.ParseBool(os.Getenv("NDT_OMIT_DELTAS"))
> ```

Non-blocking question: Is there a way to give a default value so that a bad parse still leaves a good value in there? Or is that handled by the fact that false is the zero value for the boolean type?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/302)*

pboothe:
:lgtm: with one clarifying question

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/302#-:-KyvJb35T4orTnZgelEg:bgq96e9)*

gfr10598:
*[parser/ndt.go, line 25 at r3](https://reviewable.io:443/reviews/m-lab/etl/302#-KyvJAGlO24Hqn0SnXU5:-KyvMtGG6LvFu_9HEmIk:b2ygc9b) ([raw file](https://github.com/m-lab/etl/blob/ce13e46d7bf777cac03f93c942c068fb3ba7400c/parser/ndt.go#L25)):*

Non-blocking question: Is there a way to give a default value so that a bad parse still leaves a good value in there? Or is that handled by the fact that false is the zero value for the boolean type?
</blockquote></details>

Yes false is the default if the parse fails.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/302)*

PR #363, opened 2017-12-15
Dedup2 preliminary review

gfr10598:
This isn't quite ready for full review, but would like feedback before the weekend.
Please focus on code in top level dedup directory.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/363)


stephen-soltesz:
Added some high-ish level comments. Looks like a good package design.

---

Review status: 0 of 6 files reviewed at latest revision, 10 unresolved discussions.

---

*[dedup/dedup.go, line 32 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QKcn5oZJkqikCIO2m:-L0QKcn5oZJkqikCIO2n:bqbschz) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L32)):*
> ```Go
> func init() {
> 	// Always prepend the filename and line number.
> 	log.SetFlags(log.LstdFlags | log.Lshortfile)
> ```

I'm still not sure a package should set these flags to avoid conflicts between packages. I recommend placing this in the cmd and test files instead.

---

*[dedup/dedup.go, line 42 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QKG-fr1U2OVZtiDeW:-L0QKG-fr1U2OVZtiDeX:bo76u8z) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L42)):*
> ```Go
> const (
> 	// ErrorNotRegularTable is returned when a table is not a regular table (e.g. views)
> 	ErrorNotRegularTable = Error("Not a regular table")
> ```

Out of curiosity, why a custom error type and not `fmt.Errorf`, or `errors.New`?

---

*[dedup/dedup.go, line 52 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QD-XPtqB_2qW92lPu:-L0QD-XPtqB_2qW92lPv:bjeynps) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L52)):*
> ```Go
> 
> // Detail provides more detailed information about a partition.
> type Detail struct {
> ```

s/Detail/PartitionDetail/ ?

---

*[dedup/dedup.go, line 61 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QHZEHmqC0aTmUhshP:-L0QHZEHmqC0aTmUhshQ:bbxhtp7) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L61)):*
> ```Go
> // Expects table to have test_id, task_filename, and parse_time fields.
> // `partition` should be in YYYY-MM-DD format.
> func GetNDTTableDetail(dsExt *bqext.Dataset, table string, partition string) (Detail, error) {
> ```

What makes this NDT specific?

---

*[dedup/dedup.go, line 72 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QCTH1klt0TdA0Ffs2:-L0QCTH1klt0TdA0Ffs3:b17xr27) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L72)):*
> ```Go
> 	queryString := fmt.Sprintf(`
> 		#standardSQL
> 		select sum(Tests) as TestCount, count(Task)-1 as TaskFileCount
> ```

I suggest that we adopt consistent style for SQL queries. In this case, I recommend capitalizing SQL keywords throughout.

---

*[dedup/dedup.go, line 78 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QClQUxx65hvlQcc_D:-L0QClQUxx65hvlQcc_E:b-jsch1v) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L78)):*
> ```Go
>   		    union all
> 		  	select count(test_id) as Tests, task_filename as Task
> 		  	from `+"`"+"%s"+"`"+`
> ```

This format concatenation is not easy to parse. Is there any clearer way to express it?

---

*[dedup/dedup.go, line 157 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QG2LfFAO0UNmME1G3:-L0QG2LfFAO0UNmME1G4:b-htuwel) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L157)):*
> ```Go
> // minSrcAge
> // force - if true, will ignore age and test count sanity checks
> func CheckAndDedup(dsExt *bqext.Dataset, srcInfo TableInfo, destDataset, destBase string, minSrcAge time.Duration, force bool) (bool, error) {
> ```

This is a pretty large function. It looks like there are three logical steps: setup dest table, run checks, run dedup. Could the first two be separate functions? Then the implementation of CheckAndDedup becomes shorter with clearer flow.

e.g.
```
destTable, destInfo, err := SetupDestTable(...)
if err ...
if !force {
    err := Check(destTable, destInfo)
    if err ...
}
Dedup(...)
```

---

*[dedup/dedup.go, line 276 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QIhjFu0o0fFNPbBar:-L0QIhjFu0o0fFNPbBas:b-skuc9u) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L276)):*
> ```Go
> 		match := srcTemplateRE.FindStringSubmatch(srcInfo.Name)
> 		if len(match) != 5 {
> 			log.Println("Skipping ", srcInfo.Name)
> ```

"Skipping" wants a `continue`?

---

*[dedup/dedup.go, line 278 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QIwkS7fC-O0h5ckNN:-L0QIwkS7fC-O0h5ckNO:b-6dafry) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L278)):*
> ```Go
> 			log.Println("Skipping ", srcInfo.Name)
> 		}
> 		destDate := fmt.Sprintf("%s-%s-%s", match[2], match[3], match[4])
> ```

This date formatting occurs in three places. Would a helper func help?

---

*[dedup/dedup_integration_test.go, line 19 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QJEWraGYiLci6D2CZ:-L0QJEWsZRw7LQMvrhWq:b-2gngqm) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup_integration_test.go#L19)):*
> ```Go
> 	opts := []option.ClientOption{}
> 	if os.Getenv("TRAVIS") != "" {
> 		authOpt := option.WithCredentialsFile("../travis-testing.key")
> ```

Perhaps reference issue -- https://github.com/m-lab/go/issues/11

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/363#-:-L0QKuWl5j1A4A1XRqfQ:b-cm5fo9)*

bryanbloom531:
K

gfr10598:
Review status: 0 of 6 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[dedup/dedup.go, line 32 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QKcn5oZJkqikCIO2m:-L0eEvEq8k0xslQbDNpW:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L32)):*

I'm still not sure a package should set these flags to avoid conflicts between packages. I recommend placing this in the cmd and test files instead.
</blockquote></details>

Done.

---

*[dedup/dedup.go, line 42 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QKG-fr1U2OVZtiDeW:-L0QYI9HRnuPQuhr6yqJ:bvhgcs5) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L42)):*

Out of curiosity, why a custom error type and not `fmt.Errorf`, or `errors.New`?
</blockquote></details>

So that there is a static definition that calling code can compare or switch against.
I try to make constants when I expect to actually see the error, and don't bother if the error isn't expected to happen, or should happen rarely.

---

*[dedup/dedup.go, line 52 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QD-XPtqB_2qW92lPu:-L0eFDVuO582yYmzgKbY:b-7y1rmg) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L52)):*

s/Detail/PartitionDetail/ ?
</blockquote></details>

On the fence.  It is used both for partitions, and for tables (ndt_2017...).

---

*[dedup/dedup.go, line 61 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QHZEHmqC0aTmUhshP:-L0eFboFqZiAmNzDdKeU:b-oebtsw) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L61)):*

What makes this NDT specific?
</blockquote></details>

Really isn't now.  Changed name.

---

*[dedup/dedup.go, line 72 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QCTH1klt0TdA0Ffs2:-L0eGHwWur4Bj-94oL6z:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L72)):*

I suggest that we adopt consistent style for SQL queries. In this case, I recommend capitalizing SQL keywords throughout.
</blockquote></details>

Done.

---

*[dedup/dedup.go, line 78 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QClQUxx65hvlQcc_D:-L0eGrzoHPzF5mUEsC1i:b-d5iqw4) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L78)):*

This format concatenation is not easy to parse. Is there any clearer way to express it?
</blockquote></details>

Eliminated some "s.  Should be better now.

---

*[dedup/dedup.go, line 157 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QG2LfFAO0UNmME1G3:-L0elBdiJ66Lerc-krDZ:b4k1g8f) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L157)):*

This is a pretty large function. It looks like there are three logical steps: setup dest table, run checks, run dedup. Could the first two be separate functions? Then the implementation of CheckAndDedup becomes shorter with clearer flow.

e.g.
```
destTable, destInfo, err := SetupDestTable(...)
if err ...
if !force {
    err := Check(destTable, destInfo)
    if err ...
}
Dedup(...)
```
</blockquote></details>

Took a fair amount of refactoring, but its done.

---

*[dedup/dedup.go, line 276 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QIhjFu0o0fFNPbBar:-L0eEfxSfDanBEgF8Au6:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L276)):*

"Skipping" wants a `continue`?
</blockquote></details>

Done.

---

*[dedup/dedup.go, line 278 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QIwkS7fC-O0h5ckNN:-L0eE_ixJfjeZElAw3iz:b-b3jbv4) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L278)):*

This date formatting occurs in three places. Would a helper func help?
</blockquote></details>

Obsolete.

---

*[dedup/dedup_integration_test.go, line 19 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QJEWraGYiLci6D2CZ:-L0eoJGNXfXJBZZHDB5g:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup_integration_test.go#L19)):*

Perhaps reference issue -- https://github.com/m-lab/go/issues/11
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/363)*

gfr10598:
PTAL.

---

Review status: 0 of 6 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/363#-:-L0eok-P7PyKK8vOvf_F:b-83330m)*

stephen-soltesz:
Reviewed 1 of 4 files at r7.
Review status: 1 of 6 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[dedup/dedup.go, line 42 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QKG-fr1U2OVZtiDeW:-L0eq_-DOwSoUu71lV4P:b4tdw4k) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L42)):*

So that there is a static definition that calling code can compare or switch against.
I try to make constants when I expect to actually see the error, and don't bother if the error isn't expected to happen, or should happen rarely.
</blockquote></details>

Using the built-in error type allows for comparisons also. See for example https://golang.org/pkg/io/#pkg-variables and https://godoc.org/google.golang.org/api/iterator#pkg-variables

I think using the built-in type is preferable when it is possible. Are there any use cases that requires a custom type?

---

*[dedup/dedup.go, line 72 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QCTH1klt0TdA0Ffs2:-L0eracmwGvRJF-zzWVM:behwhet) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L72)):*

Done.
</blockquote></details>

Please capitalize all SQL keywords. See also, `SUM`, `AS`, `FROM`.

---

*[dedup/dedup.go, line 130 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0f9dR3KwpX8VQ22DPd:-L0f9dR3KwpX8VQ22DPe:b-klzruz) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L130)):*
> ```Go
> // for a single table partition.
> // Returns slice ordered by decreasing age.
> func GetInfoMatching(dsExt *bqext.Dataset, filter string) ([]TableInfo, error) {
> ```

My thought is that this is another variation of `GetTableInfo`. As I've suggested in other cases, by sharing more of the names between these two functions, their relationship is more obvious. For example, if this function were name, `GetTableInfoMatching` a new reader could reasonably and immediately expect that it would have the same semantics as `GetTableInfo`.

---

*[dedup/dedup.go, line 182 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0etlFVEEZTD8IZc2D9:-L0etlFVEEZTD8IZc2DA:b-pgrsh6) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L182)):*
> ```Go
> 		`#legacySQL
> 		SELECT
> 		  partition_id as PartitionID,
> ```

nit: s/as/AS/

---

*[dedup/dedup.go, line 193 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0euVJUPSGxU9YAOc7v:-L0euVJVhqu9iOGeiaOI:b1pahb8) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L193)):*
> ```Go
> 	if err != nil {
> 		// If the partition doesn't exist, just return empty Info, no error.
> 		if err.Error() == "no more items in iterator" {
> ```

Is there a way to use static error types here? Comparing to error strings looks fragile.

e.g.
```
if err == iterator.Done {
   ...
}
```

---

*[dedup/dedup.go, line 256 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0fGihU2mGViSdnFGET:-L0fGihU2mGViSdnFGEU:b-d2g38) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L256)):*
> ```Go
> 	// If the source table is older than the destination table, then
> 	// we probably don't want to overwrite it.
> 	if srcInfo.LastModifiedTime.Before(destInfo.LastModifiedTime) {
> ```

I don't understand why we have to check the dest partition and the destination table. (not saying it's incorrect, just that I don't understand). 

Under what circumstances could the dst table last modified time be earlier than src but the dest partition info later?

Without knowing the details, I imagine writing to a table partition would update the table lastmodifiedtime. Is that incorrect?

---

*[dedup/dedup.go, line 275 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0fAO86b13Y8l2VKjNV:-L0fAO86b13Y8l2VKjNW:b-bv1mnn) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L275)):*
> ```Go
> // ignoreDestAge - if true, will ignore destination age sanity check
> func CheckAndDedup(dsExt *bqext.Dataset, srcInfo TableInfo, destTable *bigquery.Table, minSrcAge time.Duration, ignoreDestAge bool) (bool, error) {
> 	// Check if the last update was at least fDelay in the past.
> ```

fdelay does not exist.

---

*[dedup/dedup.go, line 323 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0fBHWukrzJ37aDqJ9U:-L0fBHWukrzJ37aDqJ9V:b-rw8lv5) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L323)):*
> ```Go
> 		srcInfo := info[i]
> 		// Skip any partition that has been updated in the past 48 hours.
> 		if srcInfo.LastModifiedTime.After(time.Now().Add(minAge)) {
> ```

Will minAge be negative? `time.Now().Add()` looks like it will always be a future date, which the `LastModifiedTime` could never be after. I'm looking through this quickly, so I may be missing something.

Also, it looks like the comment should reference minAge rather than a constant.

---

*[dedup/dedup.go, line 326 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0fCqGhWWIHwiWo1W57:-L0fCqGhWWIHwiWo1W58:b-yu4bjg) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L326)):*
> ```Go
> 			continue
> 		}
> 		_, err := GetTableDetail(dsExt, dsExt.Table(srcInfo.Name))
> ```

Is this check necessary? It's unclear why since ` checkTasksAndTest` eventually runs `GetTableDetail on the src table. And, `CheckAndDedup` also checks that the src table exists? Again, I'm looking through this quickly, so I may be missing something.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/363)*

gfr10598:
Review status: 1 of 6 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[dedup/dedup.go, line 42 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QKG-fr1U2OVZtiDeW:-L0fj8ayxnt0BlHpMbBJ:btwh0mi) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L42)):*

Using the built-in error type allows for comparisons also. See for example https://golang.org/pkg/io/#pkg-variables and https://godoc.org/google.golang.org/api/iterator#pkg-variables

I think using the built-in type is preferable when it is possible. Are there any use cases that requires a custom type?
</blockquote></details>

Got it.  Misinterpreted the question.
I was using custom so that I could make constants, but not worthwhile.  Changed.

---

*[dedup/dedup.go, line 72 at r5](https://reviewable.io:443/reviews/m-lab/etl/363#-L0QCTH1klt0TdA0Ffs2:-L0fjMaY9Sz9IyTiciQs:bch9e85) ([raw file](https://github.com/m-lab/etl/blob/16cf6d5f7410bf43c0d0aebc7a7d96b3fefc73ac/dedup/dedup.go#L72)):*

Please capitalize all SQL keywords. See also, `SUM`, `AS`, `FROM`.
</blockquote></details>

Done.  (Oops)

---

*[dedup/dedup.go, line 130 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0f9dR3KwpX8VQ22DPd:-L0fjwKEYuH6HwQrdJb0:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L130)):*

My thought is that this is another variation of `GetTableInfo`. As I've suggested in other cases, by sharing more of the names between these two functions, their relationship is more obvious. For example, if this function were name, `GetTableInfoMatching` a new reader could reasonably and immediately expect that it would have the same semantics as `GetTableInfo`.
</blockquote></details>

Done.

---

*[dedup/dedup.go, line 182 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0etlFVEEZTD8IZc2D9:-L0fkBfiUI91LQPynAh-:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L182)):*

nit: s/as/AS/
</blockquote></details>

Done.

---

*[dedup/dedup.go, line 193 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0euVJUPSGxU9YAOc7v:-L0fkvDLgvUvbXfrtrO2:b-n2dhy6) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L193)):*

Is there a way to use static error types here? Comparing to error strings looks fragile.

e.g.
```
if err == iterator.Done {
   ...
}
```
</blockquote></details>

Found it.  iterator.Done

---

*[dedup/dedup.go, line 256 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0fGihU2mGViSdnFGET:-L0fnSrYdPJ31a0RcSgx:b-ekezl8) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L256)):*

I don't understand why we have to check the dest partition and the destination table. (not saying it's incorrect, just that I don't understand). 

Under what circumstances could the dst table last modified time be earlier than src but the dest partition info later?

Without knowing the details, I imagine writing to a table partition would update the table lastmodifiedtime. Is that incorrect?
</blockquote></details>

Not sure why I checked both.  Removed one.

---

*[dedup/dedup.go, line 275 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0fAO86b13Y8l2VKjNV:-L0flHxB3oUhKMyaDqQ9:b32w6h) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L275)):*

fdelay does not exist.
</blockquote></details>

Tx

---

*[dedup/dedup.go, line 323 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0fBHWukrzJ37aDqJ9U:-L0flch4OFBBZiylAHtQ:b-z82v2k) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L323)):*

Will minAge be negative? `time.Now().Add()` looks like it will always be a future date, which the `LastModifiedTime` could never be after. I'm looking through this quickly, so I may be missing something.

Also, it looks like the comment should reference minAge rather than a constant.
</blockquote></details>

Good catch.  Thanks.

---

*[dedup/dedup.go, line 326 at r7](https://reviewable.io:443/reviews/m-lab/etl/363#-L0fCqGhWWIHwiWo1W57:-L0fm1NFoPNKGUrIvTHS:b-bidn6w) ([raw file](https://github.com/m-lab/etl/blob/3f0bd45687c35f215c0b64fbf003a56bc02f599e/dedup/dedup.go#L326)):*

Is this check necessary? It's unclear why since ` checkTasksAndTest` eventually runs `GetTableDetail on the src table. And, `CheckAndDedup` also checks that the src table exists? Again, I'm looking through this quickly, so I may be missing something.
</blockquote></details>

Yup.  Didn't delete this when I moved the check into CheckAndDedup several days ago.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/363)*

PR #375, opened 2018-01-03
Adjust insert buffer size for no-deltas

gfr10598:
When NDT_OMIT_DELTAS is set, each row is much smaller, so we can buffer more rows.  This should improve throughput, and likely reduce the number of workers or instances needed.

I'm not entirely happy about having NDT details in globals.go, but it seems much better than having them in insert.go.   If you have a better idea please suggest.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/375)

### DIFF
--- a/bq/insert.go
+++ b/bq/insert.go
@@ -47,7 +47,7 @@ func NewInserter(dataset string, dt etl.DataType, partition time.Time) (etl.Inse
 
 	return NewBQInserter(
 		etl.InserterParams{Dataset: dataset, Table: table, Suffix: suffix,
-			Timeout: 15 * time.Minute, BufferSize: etl.DataTypeToBQBufferSize[dt], RetryDelay: 30 * time.Second},
+			Timeout: 15 * time.Minute, BufferSize: dt.BQBufferSize(), RetryDelay: 30 * time.Second},
 		nil)
 }
 
--- a/cmd/etl_worker/app-ndt-batch.yaml
+++ b/cmd/etl_worker/app-ndt-batch.yaml
@@ -42,6 +42,7 @@ network:
     - 9090/tcp
 
 env_variables:
+  NDT_BATCH: 'true'  # Allow instances to discover they are NDT_BATCH instances.
   MAX_WORKERS: 20
   BIGQUERY_PROJECT: 'measurement-lab'
   BIGQUERY_DATASET: 'batch'
--- a/cmd/reproc/reproc_test.go
+++ b/cmd/reproc/reproc_test.go
@@ -25,7 +25,7 @@ func Test_queueFor(t *testing.T) {
 	defer ResetFlags()
 
 	tme, _ := time.Parse("2006/01/02", "2017/09/01")
-	queue := queueFor(tme)
+	queue := queueForDate(tme)
 	if queue != "base-2" {
 		t.Error("bad queue: ", queue)
 	}
--- a/etl/globals.go
+++ b/etl/globals.go
@@ -28,6 +28,8 @@ var (
 	podPattern   = regexp.MustCompile(mlabN_podNN)
 )
 
+
+
 type DataPath struct {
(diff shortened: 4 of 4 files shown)


pboothe:
A bug and two comments.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[etl/globals.go, line 88 at r2](https://reviewable.io:443/reviews/m-lab/etl/375#-L1w_I4efbK3SUehti83:-L1w_I4efbK3SUehti84:bd68x1j) ([raw file](https://github.com/m-lab/etl/blob/308d24c593de4613969ff61b0519230a9d2490b4/etl/globals.go#L88)):*
> ```Go
> 		}
> 	}
> 	return 5 * DataTypeToBQBufferSize[dt]
> ```

This returns `5 * DataTypeToBQBufferSize[dt]` on all codepaths which I think is an error?

Also, 5 is an annoying magic number. Can we instead have a special entry in  `DataTypeToBQBufferSize` which handles this case, and then the innermost `if` becomes
```go
if omitDeltas {
    return DataTypeToBQBufferSize[NDT_OMIT_DELTAS]
}
```
and then all our magic numbers are in the `DataTypeToBQBufferSize` map.

---

*[etl/globals.go, line 121 at r2](https://reviewable.io:443/reviews/m-lab/etl/375#-L1wb2cuVUWufpBNFBt5:-L1wb2cuVUWufpBNFBt6:b4g7ctb) ([raw file](https://github.com/m-lab/etl/blob/308d24c593de4613969ff61b0519230a9d2490b4/etl/globals.go#L121)):*
> ```Go
> 	// Map from data type to number of buffer size for BQ insertion.
> 	// TODO - this should be loaded from a config.
> 	DataTypeToBQBufferSize = map[DataType]int{
> ```

Should this be lower-case because it no longer needs to be exported?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/375#-:-L1wbIs5nmXX_KwZjuIB:b-4zl4sz)*

gfr10598:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[etl/globals.go, line 88 at r2](https://reviewable.io:443/reviews/m-lab/etl/375#-L1w_I4efbK3SUehti83:-L1wmwdDbu0uIUNzZIdT:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/308d24c593de4613969ff61b0519230a9d2490b4/etl/globals.go#L88)):*

This returns `5 * DataTypeToBQBufferSize[dt]` on all codepaths which I think is an error?

Also, 5 is an annoying magic number. Can we instead have a special entry in  `DataTypeToBQBufferSize` which handles this case, and then the innermost `if` becomes
```go
if omitDeltas {
    return DataTypeToBQBufferSize[NDT_OMIT_DELTAS]
}
```
and then all our magic numbers are in the `DataTypeToBQBufferSize` map.
</blockquote></details>

Done.

---

*[etl/globals.go, line 121 at r2](https://reviewable.io:443/reviews/m-lab/etl/375#-L1wb2cuVUWufpBNFBt5:-L1wmu6inZ-PCiZKgWja:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/308d24c593de4613969ff61b0519230a9d2490b4/etl/globals.go#L121)):*

Should this be lower-case because it no longer needs to be exported?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/375)*

pboothe:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/375#-:-L1wpBDJ1mIZTv1eQOJG:bnfp4nl)*

PR #377, opened 2018-01-05
Add metrics for PT tests

yachang:
Calculate number of tests reaching the expected dest_IP or not per site.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/377)


gfr10598:
I don't think it is worthwhile adding the site name, as this will generate large prometheus footprint.  Perhaps add just the metro prefix, lga, sea, ...

Also, I suggest just counting the unexpected results, so you don't need 'kind' - instead just call the counter PTTraceDestNotReached, and have the single site parameter.

BTW, I guess there are also cases where the destination is reached, but there is additional garbage added on the end.  Is it easy to count and tag these as well?  Once we use repeated fields for PT, we could add an anomalies struct, as in NDT, and have flags for DestNotReached and ExtraneousHops that could then be used for filtering.

---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[parser/pt.go, line 198 at r1](https://reviewable.io:443/reviews/m-lab/etl/377#-L26PXphtGffo5r2vGzE:-L26PXphtGffo5r2vGzF:b-4bbweg) ([raw file](https://github.com/m-lab/etl/blob/f57fc85c097db6c60fa7e48048ee10f28ec3060e/parser/pt.go#L198)):*
> ```Go
> // Extract site name like "mlab1.lga06" from file name like
> // 20170501T000000Z-mlab1-acc02-paris-traceroute-0000.tgz
> func GetSiteName(raw_fn string) string {
> ```

This should be common code in etl package I think.  We are already using code there to parse the taskfile name.

---

*[parser/pt.go, line 451 at r1](https://reviewable.io:443/reviews/m-lab/etl/377#-L26Pjtji7a67qft8t-n:-L26Pjtji7a67qft8t-o:be8lc64) ([raw file](https://github.com/m-lab/etl/blob/f57fc85c097db6c60fa7e48048ee10f28ec3060e/parser/pt.go#L451)):*
> ```Go
> 	} else {
> 		metrics.PTReachDestCount.WithLabelValues(
> 			siteName, "PT test reach expected destination IP!").Inc()
> ```

Drop this. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/377#-:-L26NtkBhVChPzTmPo09:b-l9i8oq)*

gfr10598:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[metrics/metrics.go, line 33 at r2](https://reviewable.io:443/reviews/m-lab/etl/377#-L2796I7V0vpWnQw4jz8:-L2796I8QQJ7Tn1k6AH8:bah5a17) ([raw file](https://github.com/m-lab/etl/blob/607fa97ec97118ea7ed0152890661ae52b9b1c0a/metrics/metrics.go#L33)):*
> ```Go
> 	prometheus.MustRegister(WarningCount)
> 	prometheus.MustRegister(PTNotReachDestCount)
> 	prometheus.MustRegister(PTTestCountPerSite)
> ```

Just PTTestCount.  Site is implicit in the vector label.

Please move these 2 lines up with PTHopCount

---

*[metrics/metrics.go, line 208 at r2](https://reviewable.io:443/reviews/m-lab/etl/377#-L279Q1rYCL9-ymhwoIV:-L279Q1rYCL9-ymhwoIW:b-b50pwf) ([raw file](https://github.com/m-lab/etl/blob/607fa97ec97118ea7ed0152890661ae52b9b1c0a/metrics/metrics.go#L208)):*
> ```Go
> 			Help: "Count how many PT tests did not reach expected destination per site.",
> 		},
> 		// sea03
> ```

Hmm.  I meant metro, rather than site.  I think that will be more manageable.
Stephen reminds me that appengine also adds service, version and instance labels, so we need to be more careful in appengine about adding more dimensions.

Since we will also be able to get more detail eventually from bigquery, we don't need that detail here.  So I think the three character metro code is a good level of granularity for this.

---

*[metrics/metrics.go, line 224 at r2](https://reviewable.io:443/reviews/m-lab/etl/377#-L27AXXyxLpOufbzAfH6:-L27AXXzGAy3dwveJvms:b-4rw227) ([raw file](https://github.com/m-lab/etl/blob/607fa97ec97118ea7ed0152890661ae52b9b1c0a/metrics/metrics.go#L224)):*
> ```Go
> 		},
> 		// sea03
> 		[]string{"site"},
> ```

same here - metro instead of site.

---

*[parser/pt.go, line 198 at r1](https://reviewable.io:443/reviews/m-lab/etl/377#-L26PXphtGffo5r2vGzE:-L27AbDi__LzttybKBBt:bhaqm8x) ([raw file](https://github.com/m-lab/etl/blob/f57fc85c097db6c60fa7e48048ee10f28ec3060e/parser/pt.go#L198)):*

This should be common code in etl package I think.  We are already using code there to parse the taskfile name.
</blockquote></details>

Please use existing functionality in etl package, or move this to that package.  Prefer to use the existing regexp defined in etl.

See use of ValidateTestPath in ndt.go

I will start passing in the full DataPath struct sometime soon, which will make life a little easier.  In the short term, generate it yourself as in ndt.go

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/377)*

yachang:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[metrics/metrics.go, line 33 at r2](https://reviewable.io:443/reviews/m-lab/etl/377#-L2796I7V0vpWnQw4jz8:-L2LYIL3BA5HNcfGw1tQ:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/607fa97ec97118ea7ed0152890661ae52b9b1c0a/metrics/metrics.go#L33)):*

Just PTTestCount.  Site is implicit in the vector label.

Please move these 2 lines up with PTHopCount
</blockquote></details>

Done.

---

*[metrics/metrics.go, line 208 at r2](https://reviewable.io:443/reviews/m-lab/etl/377#-L279Q1rYCL9-ymhwoIV:-L2LZXsExcUvKOo0Fx9w:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/607fa97ec97118ea7ed0152890661ae52b9b1c0a/metrics/metrics.go#L208)):*

Hmm.  I meant metro, rather than site.  I think that will be more manageable.
Stephen reminds me that appengine also adds service, version and instance labels, so we need to be more careful in appengine about adding more dimensions.

Since we will also be able to get more detail eventually from bigquery, we don't need that detail here.  So I think the three character metro code is a good level of granularity for this.
</blockquote></details>

Done.

---

*[metrics/metrics.go, line 224 at r2](https://reviewable.io:443/reviews/m-lab/etl/377#-L27AXXyxLpOufbzAfH6:-L2LZYe6kgraonkqAsVQ:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/607fa97ec97118ea7ed0152890661ae52b9b1c0a/metrics/metrics.go#L224)):*

same here - metro instead of site.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/377)*

yachang:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[parser/pt.go, line 198 at r1](https://reviewable.io:443/reviews/m-lab/etl/377#-L26PXphtGffo5r2vGzE:-L2LtswJZBuzRpRwZyg6:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/f57fc85c097db6c60fa7e48048ee10f28ec3060e/parser/pt.go#L198)):*

Please use existing functionality in etl package, or move this to that package.  Prefer to use the existing regexp defined in etl.

See use of ValidateTestPath in ndt.go

I will start passing in the full DataPath struct sometime soon, which will make life a little easier.  In the short term, generate it yourself as in ndt.go

</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/377)*

yachang:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[parser/pt.go, line 451 at r1](https://reviewable.io:443/reviews/m-lab/etl/377#-L26Pjtji7a67qft8t-n:-L2LvdAgicZ8tNPtEiVC:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/f57fc85c097db6c60fa7e48048ee10f28ec3060e/parser/pt.go#L451)):*

Drop this. 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/377)*

yachang:
All fixed. Thanks!

---

Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/377#-:-L2LvyPFhrkfQ8KrC4jV:b-qtt00a)*

gfr10598:
One more change, then I think its done!

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[etl/globals.go, line 77 at r6](https://reviewable.io:443/reviews/m-lab/etl/377#-L2LykxlgaJmfHkBr26W:-L2LykxlgaJmfHkBr26X:b-wgy46n) ([raw file](https://github.com/m-lab/etl/blob/b7c980e2ea6babc2a402721aa43cf327fd42b3d5/etl/globals.go#L77)):*
> ```Go
> // 20170501T000000Z-mlab1-acc02-paris-traceroute-0000.tgz
> func GetMetroName(raw_fn string) string {
> 	mlabN_podNN := regexp.MustCompile(`-(mlab\d)-([[:alpha:]]{3}\d[0-9t])-`)
> ```

Please use podPattern, which is already defined above.  Should only call MustCompile from init or global variable declarations, or else catch the panic that might result.  Otherwise, if a typo makes it into production, it might crash when it gets a request, instead of at start up.

Even better, use TaskPattern, or better yet, use ValidateTestPath, and then pull the Pod field out of the returned DataPath struct.  That way you will inherit any updates to the task filename parsing code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/377#-:-L2M-Yf5NS7RjP9I7OeO:bg9wltb)*

yachang:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[etl/globals.go, line 77 at r6](https://reviewable.io:443/reviews/m-lab/etl/377#-L2LykxlgaJmfHkBr26W:-L2M0N2KHgbFymhFdfOQ:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/b7c980e2ea6babc2a402721aa43cf327fd42b3d5/etl/globals.go#L77)):*

Please use podPattern, which is already defined above.  Should only call MustCompile from init or global variable declarations, or else catch the panic that might result.  Otherwise, if a typo makes it into production, it might crash when it gets a request, instead of at start up.

Even better, use TaskPattern, or better yet, use ValidateTestPath, and then pull the Pod field out of the returned DataPath struct.  That way you will inherit any updates to the task filename parsing code.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/377)*

gfr10598:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/377#-:-L2M0nTW92WATq_rh7cR:bnfp4nl)*

PR #390, opened 2018-01-19
Queue tweak

gfr10598:
Adjust batch queue rates.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/390)


pboothe:
I remain suspicious of this manual scaling approach, but :LGTM: - I think it will be a few years before we have to make it scale automatically, and the automatic scaling supported by GCP should be better-designed by then.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/390#-:-L3FVeLiggkD0vaZ9h_U:b-cwgshe)*

PR #415, opened 2018-02-06
Add "table" label to all common etl metrics

stephen-soltesz:
This change adds a new "table" label to all common ETL parser metrics.

As well, the file `etl_test.go` is renamed to `globals_test.go` to match the location of function definitions tested in `globals_test.go`. There are new tests to improve the code coverage of globals.go.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/415)

PR #425, opened 2018-02-06
Gardener initial service definition

gfr10598:
This introduces the skeleton for the gardener service, and travis deployment rules for the sandbox and staging instances.

Next steps will introduce tasks that run periodically to find the oldest partitions in the target table, and to find candidates for deduplication.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/425)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -152,7 +152,7 @@ deploy:
     # These injected params specify the project and bucket used by the service.
     INJECTED_PROJECT=mlab-sandbox INJECTED_BUCKET=archive-mlab-sandbox
     $TRAVIS_BUILD_DIR/travis/deploy_app.sh mlab-sandbox
-    /tmp/gardener.json $TRAVIS_BUILD_DIR/cmd/gardener app.yaml
+    /tmp/mlab-sandbox.json $TRAVIS_BUILD_DIR/cmd/gardener app.yaml
   skip_cleanup: true
   on:
     repo: m-lab/etl
@@ -167,7 +167,7 @@ deploy:
     # These injected params specify the project and bucket used by the service.
     INJECTED_PROJECT=mlab-staging INJECTED_BUCKET=archive-mlab-staging
     $TRAVIS_BUILD_DIR/travis/deploy_app.sh mlab-staging
-    /tmp/gardener.json $TRAVIS_BUILD_DIR/cmd/gardener app.yaml
+    /tmp/mlab-staging.json $TRAVIS_BUILD_DIR/cmd/gardener app.yaml
   skip_cleanup: true
   on:
     repo: m-lab/etl
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ script:
 # Currently skipping storage tests, because they depend on GCS, and there is
 # no emulator.
 # TODO - separate storage tests into integration and lightweight.
-- MODULES="annotation appengine bq dedup etl metrics parser schema storage task web100"
+- MODULES="annotation appengine batch bq dedup etl metrics parser schema storage task web100"
 - for module in $MODULES; do
     COVER_PKGS=${COVER_PKGS}./$module/..., ;
   done
@@ -68,7 +68,7 @@ script:
 # we lose some coverage.  The corresponding cov files are overwritten, but that is OK since
 # the non-integration tests are repeated.  If we change the unit tests to NOT run when integration
 # test tag is set, then we would need to have separate cov files.
-- for module in dedup ; do
+- for module in batch dedup ; do
     go test -v -coverpkg=$COVER_PKGS -coverprofile=$module.cov github.com/m-lab/etl/$module -tags=integration ;
   done
 
@@ -93,8 +93,8 @@ deploy:
 # STAGING: Automatically deploys all pipeline services on new merge to
 # integration branch.
 # Deploys all staging services: NDT, PT, SS, QUEUE_PUSHER
-# (Currently this does not include NDT_BATCH or CRON, as CRON is in
-# development, and batch is driven by CRON.  Prod include the batch
+# (Currently this does not include NDT_BATCH or GARDENER, as GARDENER is in
+# development, and batch is driven by GARDENER.  Prod include the batch
 # pipeline because we have been using it manually.)
 # NOTE:
 #  Failure in one of the deployments will terminate the deployment sequence, leaving
@@ -141,6 +141,37 @@ deploy:
     repo: m-lab/etl
     all_branches: true
     condition: $TRAVIS_TAG == prod-*
+
+######################################################################
+## Service: etl-gardener -- Service that runs various batch jobs, such as
+#                           reprocessing and dedupping.
+#
+# SANDBOX: before code review for development code in a specific branch.
+- provider: script
+  script:
+    # These injected params specify the project and bucket used by the service.
+    INJECTED_PROJECT=mlab-sandbox INJECTED_BUCKET=archive-mlab-sandbox
+    $TRAVIS_BUILD_DIR/travis/deploy_app.sh mlab-sandbox
+    /tmp/gardener.json $TRAVIS_BUILD_DIR/cmd/gardener app.yaml
+  skip_cleanup: true
+  on:
+    repo: m-lab/etl
+    all_branches: true
+    condition: $TRAVIS_BRANCH == gardener-sandbox-*
+
+#
+# INTEGRATION: Reviewed code, merged to integration branch, will trigger
+# deployment to mlab-staging.
+- provider: script
+  script:
+    # These injected params specify the project and bucket used by the service.
+    INJECTED_PROJECT=mlab-staging INJECTED_BUCKET=archive-mlab-staging
+    $TRAVIS_BUILD_DIR/travis/deploy_app.sh mlab-staging
+    /tmp/gardener.json $TRAVIS_BUILD_DIR/cmd/gardener app.yaml
+  skip_cleanup: true
+  on:
+    repo: m-lab/etl
+    branch: integration
 
 ######################################################################
 ## Service: etl-disco-parser -- AppEngine Flexible Environment.
@@ -305,18 +336,30 @@ deploy:
 
 ######################################################################
 ## Service: queue-pusher -- AppEngine Standard Environment.
-# SANDBOX: before code review for development code in a specific branch.
+# PERSONAL SANDBOX: before code review for development code in a specific branch.
 - provider: script
   script: $TRAVIS_BUILD_DIR/.personalize_deploy.sh
     $TRAVIS_BUILD_DIR/travis/deploy_app.sh mlab-sandbox
-    /tmp/mlab-sandbox.json $TRAVIS_BUILD_DIR/appengine/queue_pusher
+    /tmp/mlab-sandbox.json $TRAVIS_BUILD_DIR/appengine/queue_pusher app.yaml
   skip_cleanup: true
   on:
     repo: m-lab/etl
     # Consider all branches and match using the condition. By default
     # "all_branches" is false, and the condition is ignored.
     all_branches: true
-    condition: $TRAVIS_BRANCH == qp-sandbox-*
+    condition: $TRAVIS_BRANCH == qp-personal-*
+
+# SANDBOX: After code review and commits to master branch. Triggers when *ANY*
+# branch is tagged with qp-sandbox-*
+- provider: script
+  script: $TRAVIS_BUILD_DIR/travis/deploy_app.sh mlab-sandbox
+    /tmp/mlab-sandbox.json $TRAVIS_BUILD_DIR/appengine/queue_pusher
+  skip_cleanup: true
+  on:
+    repo: m-lab/etl
+    all_branches: true
+    tag: true
+    condition: $TRAVIS_TAG == qp-sandbox-* || $TRAVIS_BRANCH == integration
 
 # STAGING: After code review and commits to master branch. Triggers when *ANY*
 # branch is tagged with qp-staging-*
--- a/.travis.yml
+++ b/.travis.yml
@@ -336,30 +336,18 @@ deploy:
 
 ######################################################################
 ## Service: queue-pusher -- AppEngine Standard Environment.
-# PERSONAL SANDBOX: before code review for development code in a specific branch.
+# SANDBOX: before code review for development code in a specific branch.
 - provider: script
   script: $TRAVIS_BUILD_DIR/.personalize_deploy.sh
     $TRAVIS_BUILD_DIR/travis/deploy_app.sh mlab-sandbox
-    /tmp/mlab-sandbox.json $TRAVIS_BUILD_DIR/appengine/queue_pusher app.yaml
+    /tmp/mlab-sandbox.json $TRAVIS_BUILD_DIR/appengine/queue_pusher
   skip_cleanup: true
   on:
     repo: m-lab/etl
     # Consider all branches and match using the condition. By default
     # "all_branches" is false, and the condition is ignored.
     all_branches: true
-    condition: $TRAVIS_BRANCH == qp-personal-*
-
-# SANDBOX: After code review and commits to master branch. Triggers when *ANY*
-# branch is tagged with qp-sandbox-*
-- provider: script
-  script: $TRAVIS_BUILD_DIR/travis/deploy_app.sh mlab-sandbox
-    /tmp/mlab-sandbox.json $TRAVIS_BUILD_DIR/appengine/queue_pusher
-  skip_cleanup: true
-  on:
-    repo: m-lab/etl
-    all_branches: true
-    tag: true
-    condition: $TRAVIS_TAG == qp-sandbox-* || $TRAVIS_BRANCH == integration
+    condition: $TRAVIS_BRANCH == qp-sandbox-*
 
 # STAGING: After code review and commits to master branch. Triggers when *ANY*
 # branch is tagged with qp-staging-*
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,9 @@ script:
 # we lose some coverage.  The corresponding cov files are overwritten, but that is OK since
 # the non-integration tests are repeated.  If we change the unit tests to NOT run when integration
 # test tag is set, then we would need to have separate cov files.
-- for module in batch dedup ; do
+# Note: we do not run integration tests from forked PRs b/c the SA is unavailable.
+- if [[ -n "$TEST_SERVICE_ACCOUNT_mlab_testing" ]] ; then
+  for module in batch dedup ; do
     go test -v -coverpkg=$COVER_PKGS -coverprofile=$module.cov github.com/m-lab/etl/$module -tags=integration ;
   done
 
--- a/batch/batch.go
+++ b/batch/batch.go
@@ -145,13 +145,17 @@ func (q Queuer) postOneTask(queue, fn string) error {
 // postOneDay posts all items in an ObjectIterator into the appropriate queue.
 // TODO - should this delete the data from the target partition before posting tasks?
 func (q Queuer) postOneDay(wg *sync.WaitGroup, queue string, it *storage.ObjectIterator) error {
-	qpErrCount := 0
-	gcsErrCount := 0
-	fileCount := 0
 	if wg != nil {
 		defer wg.Done()
 	}
-	var err error
+
+	fileCount := 0
+	defer func() {
+		log.Println("Added ", fileCount, " tasks to ", queue)
+	}()
+
+	qpErrCount := 0
+	gcsErrCount := 0
 	for o, err := it.Next(); err != iterator.Done; o, err = it.Next() {
 		if err != nil {
 			// TODO - should this retry?
@@ -159,7 +163,7 @@ func (q Queuer) postOneDay(wg *sync.WaitGroup, queue string, it *storage.ObjectI
 			gcsErrCount++
 			if gcsErrCount > 5 {
 				log.Printf("Failed after %d files to %s.\n", fileCount, queue)
-				break
+				return err
 			}
 			continue
 		}
@@ -168,7 +172,7 @@ func (q Queuer) postOneDay(wg *sync.WaitGroup, queue string, it *storage.ObjectI
 		if err != nil {
 			log.Println(err)
 			if strings.Contains(err.Error(), "UNKNOWN_QUEUE") {
-				break
+				return err
 			}
 			if strings.Contains(err.Error(), "invalid filename") {
 				// Invalid filename is never going to get better.
@@ -184,15 +188,14 @@ func (q Queuer) postOneDay(wg *sync.WaitGroup, queue string, it *storage.ObjectI
 				qpErrCount++
 				if qpErrCount > 5 {
 					log.Printf("Failed after %d files to %s (on %s).\n", fileCount, queue, o.Name)
-					break
+					return err
 				}
 			}
 		} else {
 			fileCount++
 		}
 	}
-	log.Println("Added ", fileCount, " tasks to ", queue)
-	return err
+	return nil
 }
 
 // PostDay fetches an iterator over the objects with ndt/YYYY/MM/DD prefix,
@@ -240,20 +243,23 @@ func (q Queuer) PostMonth(prefix string) error {
 	it := q.Bucket.Objects(context.Background(), &qry)
 
 	var wg sync.WaitGroup
+	defer func() {
+		wg.Wait()
+	}()
+
 	errCount := 0
 	const maxErrors = 20
-	var err error
 	for o, err := it.Next(); err != iterator.Done; o, err = it.Next() {
 		if err != nil {
 			log.Println(err)
 			if strings.Contains(err.Error(),
 				"does not have storage.objects.list access") {
 				// Inadequate permissions
-				break
+				return err
 			}
 			if errCount++; errCount > maxErrors {
 				log.Println("Too many errors.  Breaking loop.")
-				break
+				return err
 			}
 		} else if o.Prefix != "" {
 			wg.Add(1)
@@ -262,13 +268,12 @@ func (q Queuer) PostMonth(prefix string) error {
 				log.Println(err)
 				if errCount++; errCount > maxErrors {
 					log.Println("Too many errors.  Breaking loop.")
-					break
+					return err
 				}
 			}
 		} else {
 			log.Println("Skipping: ", o.Name)
 		}
 	}
-	wg.Wait()
-	return err
+	return nil
 }
--- a/batch/batch.go
+++ b/batch/batch.go
@@ -143,6 +143,7 @@ func (q Queuer) postOneTask(queue, fn string) error {
 }
 
 // postOneDay posts all items in an ObjectIterator into the appropriate queue.
+// TODO - should this delete the data from the target partition before posting tasks?
 func (q Queuer) postOneDay(wg *sync.WaitGroup, queue string, it *storage.ObjectIterator) error {
 	qpErrCount := 0
 	gcsErrCount := 0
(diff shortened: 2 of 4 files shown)


stephen-soltesz:
Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[.travis.yml, line 351 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4l3mAj8PrtNTUKr0Ue:-L4l3mAj8PrtNTUKr0Uf:b-1kvu2r) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/.travis.yml#L351)):*
> ```YAML
>     # "all_branches" is false, and the condition is ignored.
>     all_branches: true
>     condition: $TRAVIS_BRANCH == qp-personal-*
> ```

I think it is preferable to have a consistent conventions within a file, a single repo, and ideally across repos. Please let's standardize on one convention of using "sandbox" branches for deployments to the "sandbox" project.

If you'd like to ditch the "personalized" deployments, I'd support that too. The "table label" pr https://github.com/m-lab/etl/pull/415 removes the personalize script from the parser service deployments b/c that required additional special queue configs that afaik, we never used.

---

*[batch/batch.go, line 155 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4l85jw42FC_1w82WXO:-L4l85jw42FC_1w82WXP:b5dwzwk) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/batch/batch.go#L155)):*
> ```Go
> 	}
> 	var err error
> 	for o, err := it.Next(); err != iterator.Done; o, err = it.Next() {
> ```

I see what you're going for, but I think the outer `err`is masked by the assignment in the loop?

Example: https://play.golang.org/p/UVtAC72Op3D

---

*[batch/batch.go, line 173 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4l6lq84Z5cx8kDvThl:-L4l6lq84Z5cx8kDvThm:bg7a2bc) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/batch/batch.go#L173)):*
> ```Go
> 				break
> 			}
> 			if strings.Contains(err.Error(), "invalid filename") {
> ```

These kinds of error checks always make me a little uncomfortable. While it's out of scope for this PR, how can we handle errors or error types more generally in our code?

---

*[cmd/gardener/app.yaml, line 6 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4l8K875xrbkugoVm4G:-L4l8K875xrbkugoVm4H:bg9dlv0) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/app.yaml#L6)):*
> ```YAML
> runtime: go
> env: flex
> service: etl-gardener
> ```

Thank you for naming the service with the "etl-" prefix. This will be extremely helpful in identifying the scope of it in the AppEngine console view.

Should we rename the "etl_worker" to "worker" or would you be open to renaming the "gardener" binary to "etl_gardener"?

---

*[cmd/gardener/app.yaml, line 12 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4l90MDCVFxCz-urYdn:-L4l90MDCVFxCz-urYdo:b-rumvrq) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/app.yaml#L12)):*
> ```YAML
> #  handled transparently by chrome.  For interactive use, we would need to
> #  serve a page that has the sign-in code in it, and from there, support the
> #  requests that need credentials.  Search for "integrating google sign-in".
> ```

If you're willing to support an API-like interface, we could write a simple client library that authenticates correctly. I did this with the early api work over a year ago.

---

*[cmd/gardener/app.yaml, line 14 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4l8fqC-Gvv_GCjt54r:-L4l8fqC-Gvv_GCjt54s:b-t5svcd) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/app.yaml#L14)):*
> ```YAML
> #  requests that need credentials.  Search for "integrating google sign-in".
> #
> # For now we probably should just live with a security hole (for testing),
> ```

We have an increasing number of these. Let's start a doc enumerating them so we don't lose track.

---

*[cmd/gardener/gardener.go, line 71 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4l9zhA1xaiF10X1Xhu:-L4l9zhA1xaiF10X1Xhv:b6cfwfe) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L71)):*
> ```Go
> func Status(w http.ResponseWriter, r *http.Request) {
> 	fmt.Fprintf(w, "<html><body>\n")
> 	fmt.Fprintf(w, "<p>NOTE: This is just one of potentially many instances.</p>\n")
> ```

The current config is 1 instance. Do we ever expect the gardener to run in multi-instance mode?

---

*[cmd/gardener/gardener.go, line 93 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4lAbwq-394-frfe_S4:-L4lAbwq-394-frfe_S5:bkeap8g) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L93)):*
> ```Go
> // Month initiates reprocessing of an entire month of archive data.
> // example http://localhost:8080/reproc/month?prefix=ndt/2017/09/
> // TODO - use flusher, ok := w.(http.Flusher) to send partial result updates.
> ```

Cool. I did not know about Flusher.

---

*[cmd/gardener/gardener.go, line 106 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4lB2nJ25wtUrEzw9zj:-L4lB2nJ25wtUrEzw9zk:bce6xsv) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L106)):*
> ```Go
> 	if fields == nil {
> 		log.Printf("Invalid prefix %s\n", rawPrefix)
> 		fmt.Fprintf(rwr, "Invalid prefix %s\n", rawPrefix)
> ```

These messages will be a little confusing if `rawPrefix` is empty. Maybe use `%q` format whenever possible.

---

*[cmd/gardener/gardener.go, line 115 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4lBT1r-RrHnpIQFzCE:-L4lBT1r-RrHnpIQFzCF:b-1r7p08) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L115)):*
> ```Go
> 
> func healthCheckHandler(w http.ResponseWriter, r *http.Request) {
> 	// TODO(soltesz): provide a real health check.
> ```

Since this is service-specific, please remove my name. :-)

---

*[cmd/gardener/gardener.go, line 134 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4lC70vEMPAm0bWe7Jz:-L4lC70vEMPAm0bWe7K-:b5s4ii) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L134)):*
> ```Go
> 	if err != nil {
> 		log.Println(err)
> 		log.Fatal("Required environment variables are missing or invalid.")
> ```

What happens if a flex vm container exits? Maybe enter a running but disabled state that's monitorable. e.g. don't add the handlers below.

---

*[cmd/gardener/gardener.go, line 155 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4lCYEi7be1Eu8t7T2x:-L4lCYEi7be1Eu8t7T2y:b-wmywgl) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L155)):*
> ```Go
> 	// TODO implement listing queues to determine number of queue, and change this to 0
> 	fNumQueues = flag.Int("num_queues", 8, "Number of queues.  Normally determined by listing queues.")
> 	fBucket    = flag.String("bucket", "archive-mlab-oti", "Source bucket.")
> ```

Production defaults still make me nervous. Could this be read from the environment?

---

*[cmd/gardener/gardener.go, line 169 at r1](https://beta.reviewable.io/reviews/m-lab/etl/425#-L4lCgsbAnczZeJf5tQa:-L4lCgsbAnczZeJf5tQb:b-j816yj) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L169)):*
> ```Go
> func main() {
> 	// Check if invoked as a service.
> 	cron, _ := strconv.ParseBool(os.Getenv("GARDENER_SERVICE"))
> ```

"cron"? "enabled"? "inAppEngine"?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/m-lab/etl/425)*

gfr10598:
Review status: 0 of 4 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

*[.travis.yml, line 351 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4l3mAj8PrtNTUKr0Ue:-L4lK0Qa4lL5_damf8EE:bka96rw) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/.travis.yml#L351)):*

I think it is preferable to have a consistent conventions within a file, a single repo, and ideally across repos. Please let's standardize on one convention of using "sandbox" branches for deployments to the "sandbox" project.

If you'd like to ditch the "personalized" deployments, I'd support that too. The "table label" pr https://github.com/m-lab/etl/pull/415 removes the personalize script from the parser service deployments b/c that required additional special queue configs that afaik, we never used.
</blockquote></details>

Hmm.  I don't know how this code crept in.  Reverted.

---

*[batch/batch.go, line 155 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4l85jw42FC_1w82WXO:-L4lEiaJMKvRG5bwx4IE:bc59sla) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/batch/batch.go#L155)):*
> convention
Ah thanks!  Changed approach
Let me know if you prefer that I fix the original approach.

---

*[batch/batch.go, line 173 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4l6lq84Z5cx8kDvThl:-L4lF-ESo97fOFMIfqki:bcvdkmz) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/batch/batch.go#L173)):*

These kinds of error checks always make me a little uncomfortable. While it's out of scope for this PR, how can we handle errors or error types more generally in our code?
</blockquote></details>

Yeah - trouble is we want specific errors we can compare to, but also sometimes want to be able to attach additional info to the error.  So we probably need to develop our own error type.

---

*[cmd/gardener/app.yaml, line 6 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4l8K875xrbkugoVm4G:-L4lGWnkEwUn4kUgHUsP:b-rd5a8m) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/app.yaml#L6)):*

Thank you for naming the service with the "etl-" prefix. This will be extremely helpful in identifying the scope of it in the AppEngine console view.

Should we rename the "etl_worker" to "worker" or would you be open to renaming the "gardener" binary to "etl_gardener"?
</blockquote></details>

I think I would prefer to drop the etl_ prefix.  Seems like stuttering.  Do you have a different view?

---

*[cmd/gardener/app.yaml, line 12 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4l90MDCVFxCz-urYdn:-L4lHaPui5B0XSU6WVjp:b-hcpkt6) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/app.yaml#L12)):*

If you're willing to support an API-like interface, we could write a simple client library that authenticates correctly. I did this with the early api work over a year ago.
</blockquote></details>

Let's discuss - would like to know more about that.

---

*[cmd/gardener/app.yaml, line 14 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4l8fqC-Gvv_GCjt54r:-L4lHdqCNdtCgrMu4Eq7:bcok8bi) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/app.yaml#L14)):*

We have an increasing number of these. Let's start a doc enumerating them so we don't lose track.
</blockquote></details>

https://docs.google.com/document/d/1x46aDgatVRp0XB7vAgq2VufOa6hfHG_IICsHGUcHfs0

---

*[cmd/gardener/gardener.go, line 71 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4l9zhA1xaiF10X1Xhu:-L4lHhur09iTjU6416J3:b-yl8iv5) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L71)):*

The current config is 1 instance. Do we ever expect the gardener to run in multi-instance mode?
</blockquote></details>

Will try to avoid that.  With multiple instances there will be more synchronization issues, though I can imagine it might eventually come to that.  Prefer until then to use in process parallelism for scaling.

---

*[cmd/gardener/gardener.go, line 93 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lAbwq-394-frfe_S4:-L4lK8evH3wBbhvW-3n2:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L93)):*

Cool. I did not know about Flusher.
</blockquote></details>

Done.

---

*[cmd/gardener/gardener.go, line 106 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lB2nJ25wtUrEzw9zj:-L4lIITVgr-wHoZkHp-o:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L106)):*

These messages will be a little confusing if `rawPrefix` is empty. Maybe use `%q` format whenever possible.
</blockquote></details>

Done.

---

*[cmd/gardener/gardener.go, line 115 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lBT1r-RrHnpIQFzCE:-L4lIMlO1aUbRl59t3wn:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L115)):*

Since this is service-specific, please remove my name. :-)
</blockquote></details>

Done.

---

*[cmd/gardener/gardener.go, line 134 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lC70vEMPAm0bWe7Jz:-L4lIXGx2oZzD6Pk6u9r:b-6tyswe) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L134)):*

What happens if a flex vm container exits? Maybe enter a running but disabled state that's monitorable. e.g. don't add the handlers below.
</blockquote></details>

Hmm.  Let's discuss implications.

---

*[cmd/gardener/gardener.go, line 155 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lCYEi7be1Eu8t7T2x:-L4lIeiVYJXMU-E54R19:bye7hx) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L155)):*

Production defaults still make me nervous. Could this be read from the environment?
</blockquote></details>

I think this one is OK as long as it is never used for writing.  Gardener should ideally not have permission to modify the source bucket, and should only ever read/stat.  I'll add a comment. But is it still objectionable?

---

*[cmd/gardener/gardener.go, line 169 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lCgsbAnczZeJf5tQa:-L4lJLSEd9iEKCfwB3od:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L169)):*

"cron"? "enabled"? "inAppEngine"?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/425)*

stephen-soltesz:
Reviewed 1 of 3 files at r2.
Review status: 1 of 4 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[batch/batch.go, line 247 at r3](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lbYK2hSPdM-E5SkJm:-L4lbYK2hSPdM-E5SkJn:b-4r7o9k) ([raw file](https://github.com/m-lab/etl/blob/9ad3fc548b034f578359d599a8ff29db56ccad22/batch/batch.go#L247)):*
> ```Go
> 	var wg sync.WaitGroup
> 	defer func() {
> 		wg.Wait()
> ```

Does this work? `defer wg.Wait()`

---

*[cmd/gardener/app.yaml, line 6 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4l8K875xrbkugoVm4G:-L4lOUp3cDIijTY_iJUN:bnqv8yr) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/app.yaml#L6)):*

I think I would prefer to drop the etl_ prefix.  Seems like stuttering.  Do you have a different view?
</blockquote></details>

My view is consistency is preferable. "worker" works for me.

---

*[cmd/gardener/gardener.go, line 155 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lCYEi7be1Eu8t7T2x:-L4lPgPziWdp_ezonO91:b-4anin1) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L155)):*

I think this one is OK as long as it is never used for writing.  Gardener should ideally not have permission to modify the source bucket, and should only ever read/stat.  I'll add a comment. But is it still objectionable?
</blockquote></details>

Read-only is better. How would I deploy this service with a different value for testing? (also, I'm realizing that you're writing this as both a command line tool and a service. So, the use of flags makes sense. But, atm, I don't see a way of deploying the service with alternate options.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/425)*

gfr10598:
Review status: 1 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[batch/batch.go, line 247 at r3](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lbYK2hSPdM-E5SkJm:-L4liuggH8sCfy1EipUA:bb6t5oh) ([raw file](https://github.com/m-lab/etl/blob/9ad3fc548b034f578359d599a8ff29db56ccad22/batch/batch.go#L247)):*

Does this work? `defer wg.Wait()`
</blockquote></details>

D'oh

---

*[cmd/gardener/app.yaml, line 6 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4l8K875xrbkugoVm4G:-L4lizctVBi_ryQX_-Pm:b4nvo9k) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/app.yaml#L6)):*

My view is consistency is preferable. "worker" works for me.
</blockquote></details>

I'll add a todo I think.

---

*[cmd/gardener/gardener.go, line 155 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lCYEi7be1Eu8t7T2x:-L4ljGicUSDOjSx5moeq:b-njfyrl) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L155)):*

Read-only is better. How would I deploy this service with a different value for testing? (also, I'm realizing that you're writing this as both a command line tool and a service. So, the use of flags makes sense. But, atm, I don't see a way of deploying the service with alternate options.)
</blockquote></details>

The flags are ONLY used for the command line.  For the service, the values come from environment variables.  Will clarify in comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/425)*

stephen-soltesz:
Let's discuss in person what happens when a flex vm container exits.
And, please confirm that `defer wg.Wait()` really works as intended.

With that, things :lgtm:

---

Review status: 1 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[batch/batch.go, line 247 at r3](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lbYK2hSPdM-E5SkJm:-L4lp459IaVajr-FI06k:b-a06oaw) ([raw file](https://github.com/m-lab/etl/blob/9ad3fc548b034f578359d599a8ff29db56ccad22/batch/batch.go#L247)):*

D'oh
</blockquote></details>

This was a real question, and I'm still not sure :)

This form uses the current value of var at the time of the defer statement.
```
defer fmt.Println(var)
```

Whereas this form uses the value at execution time, which means var can be updated later.
```
defer func() {
    fmt.Println(var)
}
```

I saw that `WaitGroup` was a struct type -- and I would have expected that wg would need to be a pointer since all the receiver methods are on pointers, so, I'm really not sure whether `wg` has the right values here in the short form case. Can you confirm?

---

*[cmd/gardener/gardener.go, line 155 at r1](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lCYEi7be1Eu8t7T2x:-L4lpp9DchrHpglwwOT_:bci9vyc) ([raw file](https://github.com/m-lab/etl/blob/6545a3b10cf57519dfe12ae940cc6a0161bbafa6/cmd/gardener/gardener.go#L155)):*

The flags are ONLY used for the command line.  For the service, the values come from environment variables.  Will clarify in comment.
</blockquote></details>

Oooh. Got it. I didn't piece that together before.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/425#-:-L4lqFMIxypzT3n1QH27:b2v3jam)*

gfr10598:
Review status: 1 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[batch/batch.go, line 247 at r3](https://reviewable.io:443/reviews/m-lab/etl/425#-L4lbYK2hSPdM-E5SkJm:-L4ltw6mR8cb-olRYhmP:b2zb5ll) ([raw file](https://github.com/m-lab/etl/blob/9ad3fc548b034f578359d599a8ff29db56ccad22/batch/batch.go#L247)):*

This was a real question, and I'm still not sure :)

This form uses the current value of var at the time of the defer statement.
```
defer fmt.Println(var)
```

Whereas this form uses the value at execution time, which means var can be updated later.
```
defer func() {
    fmt.Println(var)
}
```

I saw that `WaitGroup` was a struct type -- and I would have expected that wg would need to be a pointer since all the receiver methods are on pointers, so, I'm really not sure whether `wg` has the right values here in the short form case. Can you confirm?
</blockquote></details>

Wait() takes pointer receiver, so I think this should be fine.
Pretty sure the unit test would fail if this didn't work.  It does fail if the defer is commented out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/425)*

gfr10598:
PTAL.  Trying out split health check.

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/425#-:-L4mWS0vMHwASZtnRvwI:bdgrkur)*

stephen-soltesz:
:lgtm:

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[cmd/gardener/app.yaml, line 54 at r4](https://reviewable.io:443/reviews/m-lab/etl/425#-L4mY-VmEMEE9lI7HB0l:-L4mY-VmEMEE9lI7HB0m:bv54ctm) ([raw file](https://github.com/m-lab/etl/blob/04c6af024eeef6da29374eb5a8943bfb0ba0a308/cmd/gardener/app.yaml#L54)):*
> ```YAML
>     - 9090/tcp
> 
>  liveness_check:
> ```

Please add a comment like "Instances that fail liveness checks are restarted."

---

*[cmd/gardener/app.yaml, line 61 at r4](https://reviewable.io:443/reviews/m-lab/etl/425#-L4mY73X-PlPGOke9bc5:-L4mY73YKszzmGjIJT-z:bhpb083) ([raw file](https://github.com/m-lab/etl/blob/04c6af024eeef6da29374eb5a8943bfb0ba0a308/cmd/gardener/app.yaml#L61)):*
> ```YAML
>    success_threshold: 1
> 
> readiness_check:
> ```

Please add a comment like "Instances that don't pass the readiness check are not added to the pool of available instances."

And, maybe a link to https://cloud.google.com/appengine/docs/flexible/go/configuring-your-app-with-app-yaml (which is where I got these hints)

---

*[cmd/gardener/app.yaml, line 67 at r4](https://reviewable.io:443/reviews/m-lab/etl/425#-L4mYtw6-YV3WJbM01N1:-L4mYtw6-YV3WJbM01N2:bumpz1m) ([raw file](https://github.com/m-lab/etl/blob/04c6af024eeef6da29374eb5a8943bfb0ba0a308/cmd/gardener/app.yaml#L67)):*
> ```YAML
>   failure_threshold: 2
>   success_threshold: 1
>   app_start_timeout_sec: 120
> ```

I read the docs as meaning that at deployment time, "After this period, the deployment fails and is rolled back."  This is exactly what we want.

I cannot predict from the information I have what the interaction between this 2min timeout and the 30sec x 2 failure threshold from the liveness check will be. I don't know what would happen without trying it, but I'm wondering if we should  have the liveness check failure threshold be greater than the readiness start timeout?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/425#-:-L4m_L_CT-27J1nUFrjs:bnfp4nl)*

gfr10598:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[cmd/gardener/app.yaml, line 54 at r4](https://reviewable.io:443/reviews/m-lab/etl/425#-L4mY-VmEMEE9lI7HB0l:-L4pqdgMJ0PoRN6yl5KC:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/04c6af024eeef6da29374eb5a8943bfb0ba0a308/cmd/gardener/app.yaml#L54)):*

Please add a comment like "Instances that fail liveness checks are restarted."
</blockquote></details>

Done.

---

*[cmd/gardener/app.yaml, line 61 at r4](https://reviewable.io:443/reviews/m-lab/etl/425#-L4mY73X-PlPGOke9bc5:-L4ppNMGDWFRU0zH9WvU:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/04c6af024eeef6da29374eb5a8943bfb0ba0a308/cmd/gardener/app.yaml#L61)):*

Please add a comment like "Instances that don't pass the readiness check are not added to the pool of available instances."

And, maybe a link to https://cloud.google.com/appengine/docs/flexible/go/configuring-your-app-with-app-yaml (which is where I got these hints)
</blockquote></details>

Done.

---

*[cmd/gardener/app.yaml, line 67 at r4](https://reviewable.io:443/reviews/m-lab/etl/425#-L4mYtw6-YV3WJbM01N1:-L4pmvbjqmD-R0CnC9I7:bjhod6z) ([raw file](https://github.com/m-lab/etl/blob/04c6af024eeef6da29374eb5a8943bfb0ba0a308/cmd/gardener/app.yaml#L67)):*

I read the docs as meaning that at deployment time, "After this period, the deployment fails and is rolled back."  This is exactly what we want.

I cannot predict from the information I have what the interaction between this 2min timeout and the 30sec x 2 failure threshold from the liveness check will be. I don't know what would happen without trying it, but I'm wondering if we should  have the liveness check failure threshold be greater than the readiness start timeout?
</blockquote></details>

Hmm.  There is normally a 300 second period when liveness checks are ignored.  I've set that to 120, so the total delay before failing liveness will be 180.  I'll also see what the behavior looks like in a real instance.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/425)*

PR #430, opened 2018-02-07
Panic handling

gfr10598:
Any suggestions for testing?

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/430)


gfr10598:
Actually I came up with a method to allow some testing.  Will update the code.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430#-:-L4lB9K5__75DZt34aW-:b26fjyf)*

stephen-soltesz:
Reviewed 1 of 3 files at r1.
Review status: 1 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[etl/globals.go, line 202 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lDaH29JaJUb4jMzHS:-L4lDaH29JaJUb4jMzHT:b-5fcecy) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L202)):*
> ```Go
> // RunSafely executes f in a wrapper that will catch any panic
> // and return an error.
> func RunSafely(f func()) (err error) {
> ```

Hmm.

Would the following work?

`defer` statements run as a panic propagates up the stack. If we changed the prototype from `RunSafely(f)` to `CountPanic(label)`, could we change the logic to be called like:

```
func worker(rwr http.ResponseWriter, rq *http.Request) {
    defer etl.CountPanic("worker")
    .. existing code ...
}
```

My thinking is that this would make adding the panic label configurable and easy to place the counter in multiple locations.

---

*[etl/globals.go, line 212 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lDSZUc8TEpfzokcnf:-L4lDSZUc8TEpfzokcng:bsca8rj) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L212)):*
> ```Go
> 			}
> 			log.Println("Recovered from panic:", err)
> 			metrics.PanicCount.WithLabelValues("worker").Inc()
> ```

Can this label be a parameter? What if we want to run a parser or the gardener safely?

---

*[etl/globals.go, line 216 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lGEnovXwFP91bZ-WT:-L4lGEnovXwFP91bZ-WU:bamdykj) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L216)):*
> ```Go
> 	}()
> 	f()
> 	return err
> ```

Should the defer re-panic? does this effectively stop the panic from propagating?

---

*[metrics/metrics.go, line 46 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lFtz0uE7ciDO12BZg:-L4lFtz0uE7ciDO12BZh:big96z1) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/metrics/metrics.go#L46)):*
> ```Go
> 	prometheus.MustRegister(DurationHistogram)
> 	prometheus.MustRegister(InsertionHistogram)
> 	prometheus.MustRegister(FileSizeHistogram)
> ```

+ `prometheus.MustRegister(PanicCount)`

---

*[metrics/metrics.go, line 65 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lFao9ujYRhA-1jXaB:-L4lFao9ujYRhA-1jXaC:b1wdm1c) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/metrics/metrics.go#L65)):*
> ```Go
> 	PanicCount = prometheus.NewCounterVec(
> 		prometheus.CounterOpts{
> 			Name: "etl_panic_count",
> ```

https://prometheus.io/docs/practices/naming/

Please suffix counter metrics with `_total` rather than `_counter`.  I know we haven't done this consistently, but the promtool linter complains and we can work towards this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430)*

gfr10598:
Review status: 1 of 7 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[etl/globals.go, line 202 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lDaH29JaJUb4jMzHS:-L4lZUbppS8IvqeP2h6B:b-cwxlxo) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L202)):*

Hmm.

Would the following work?

`defer` statements run as a panic propagates up the stack. If we changed the prototype from `RunSafely(f)` to `CountPanic(label)`, could we change the logic to be called like:

```
func worker(rwr http.ResponseWriter, rq *http.Request) {
    defer etl.CountPanic("worker")
    .. existing code ...
}
```

My thinking is that this would make adding the panic label configurable and easy to place the counter in multiple locations.
</blockquote></details>

Well, I got something similar to work.  I think it is big improvement!

---

*[etl/globals.go, line 212 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lDSZUc8TEpfzokcnf:-L4lhiwzWzcxHghTxhJr:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L212)):*

Can this label be a parameter? What if we want to run a parser or the gardener safely?
</blockquote></details>

Done.

---

*[etl/globals.go, line 216 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lGEnovXwFP91bZ-WT:-L4lhkf6dE-YzLLjSLYj:b-f4yzhz) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L216)):*

Should the defer re-panic? does this effectively stop the panic from propagating?
</blockquote></details>

The purpose is to prevent the panic from propagating.  If that is inappropriate, then we catch the panic here.

---

*[metrics/metrics.go, line 46 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lFtz0uE7ciDO12BZg:-L4liNLH9ME5r5Uo2eDD:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/metrics/metrics.go#L46)):*

+ `prometheus.MustRegister(PanicCount)`
</blockquote></details>

Done.

---

*[metrics/metrics.go, line 65 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lFao9ujYRhA-1jXaB:-L4liO0c3S-I2-S48a9A:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/metrics/metrics.go#L65)):*

https://prometheus.io/docs/practices/naming/

Please suffix counter metrics with `_total` rather than `_counter`.  I know we haven't done this consistently, but the promtool linter complains and we can work towards this.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430)*

stephen-soltesz:
Reviewed 3 of 6 files at r3.
Review status: 4 of 7 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[etl/globals.go, line 216 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lGEnovXwFP91bZ-WT:-L4ljs03PD-P-7zjs0bs:b3s54ws) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L216)):*

The purpose is to prevent the panic from propagating.  If that is inappropriate, then we catch the panic here.
</blockquote></details>

I thought the purpose was to count the number of panics. I'm not actually sure how to reason about the state of the system if we continue executing after a panic occurs. If we count the panic, but re-propagate it then the original request will just fail like it does today, but we can alert on panic counts.

Which panics are safely recoverable? Does this cover all cases? What will be the response code seen by the taskqueue for a request with a recovered panic? Is that correct in all cases?

When debugging a system it is invaluable to be able to make concrete statements about the state of the system at various points during execution.

---

*[metrics/metrics.go, line 46 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lFtz0uE7ciDO12BZg:-L4ljOMpxESioVpAFOtO:bt9z5rn) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/metrics/metrics.go#L46)):*

Done.
</blockquote></details>

When you rebase, you'll see that these metrics are categorized slightly more. Please place PanicCount in the "common" metrics.

---

*[parser/ndt.go, line 176 at r3](https://reviewable.io:443/reviews/m-lab/etl/430#-L4ljdbhbU5e8c0YDfsK:-L4ljdbhbU5e8c0YDfsL:bt05xhy) ([raw file](https://github.com/m-lab/etl/blob/f06724314a1e6ba89494e10647d9c277940a5da6/parser/ndt.go#L176)):*
> ```Go
> 	// This will recover from any panics.
> 	defer func() {
> 		err = etl.CatchPanic(recover(), "task.ProcessAllTests")
> ```

I'm guessing this is a copy/paste bug. Please use "ndt" or "ndt.ParseAndInsert" here.

---

*[parser/ndt.go, line 176 at r3](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lmPgav6pVCZWIz2bq:-L4lmPgav6pVCZWIz2br:bzccs6a) ([raw file](https://github.com/m-lab/etl/blob/f06724314a1e6ba89494e10647d9c277940a5da6/parser/ndt.go#L176)):*
> ```Go
> 	// This will recover from any panics.
> 	defer func() {
> 		err = etl.CatchPanic(recover(), "task.ProcessAllTests")
> ```

defer statements will run even if a panic has not occurred. But `recover()` will return `nil` if there is no panic.

Does this statement work as intended when `ParseAndInsert` returns a non-nil error without a panic? Or, is `err` reset to nil? (p.s. this is an example of reasoning about the system without panic re-propagation.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430)*

yachang:
Review status: 4 of 7 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[etl/globals.go, line 202 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lquCLlBvouKrAQNZk:-L4lquCLlBvouKrAQNZl:b17vi2a) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L202)):*
> ```Go
> // RunSafely executes f in a wrapper that will catch any panic
> // and return an error.
> func RunSafely(f func()) (err error) {
> ```

Will you use RunSafely() wrapper for all parser tasks?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430)*

yachang:
Review status: 4 of 7 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[etl/globals.go, line 212 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lrgy1qP_SZ6UM-a32:-L4lrgy1qP_SZ6UM-a33:b-nb3fr7) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L212)):*
> ```Go
> 			}
> 			log.Println("Recovered from panic:", err)
> 			metrics.PanicCount.WithLabelValues("worker").Inc()
> ```

Can we make "worker" as the type of parser, and pass it as input param?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430)*

gfr10598:
Review status: 1 of 7 files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---

*[etl/globals.go, line 202 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lquCLlBvouKrAQNZk:-L4mBFKBM4xtMosgabI2:b-l81mjd) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L202)):*

Will you use RunSafely() wrapper for all parser tasks?
</blockquote></details>

RunSafely now replace with CatchPanic, which has rather different usage.

---

*[etl/globals.go, line 212 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lrgy1qP_SZ6UM-a32:-L4lvoL0gaqSCvby14V8:b14f7oj) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L212)):*

Can we make "worker" as the type of parser, and pass it as input param?
</blockquote></details>

I've replaced worker with a generic tag, generally expected to be the name of the function where the CatchPanic is located.

---

*[etl/globals.go, line 216 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lGEnovXwFP91bZ-WT:-L4lwSAKny9SfRX2XTiM:bfkaorz) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L216)):*

I thought the purpose was to count the number of panics. I'm not actually sure how to reason about the state of the system if we continue executing after a panic occurs. If we count the panic, but re-propagate it then the original request will just fail like it does today, but we can alert on panic counts.

Which panics are safely recoverable? Does this cover all cases? What will be the response code seen by the taskqueue for a request with a recovered panic? Is that correct in all cases?

When debugging a system it is invaluable to be able to make concrete statements about the state of the system at various points during execution.
</blockquote></details>

https://github.com/golang/go/wiki/PanicAndRecover asserts a convention of not allowing panics to cross package boundaries.  So the panics in Task and Parser should be captured and converted to errors.

Since etl_worker is in main, we could allow that panic to propagate and terminate that go routine.  However, I'd prefer to handle that error explicitly.

Good point about whether that error will result in the incoming task failing.  I've added explicit code to return an http error.  I believe that should provide equivalent behavior.

---

*[metrics/metrics.go, line 46 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lFtz0uE7ciDO12BZg:-L4m4V5nJ16pC5GRuPAG:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/metrics/metrics.go#L46)):*

When you rebase, you'll see that these metrics are categorized slightly more. Please place PanicCount in the "common" metrics.
</blockquote></details>

Done.

---

*[parser/ndt.go, line 176 at r3](https://reviewable.io:443/reviews/m-lab/etl/430#-L4ljdbhbU5e8c0YDfsK:-L4mB3c96_9GepdZ2CNB:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/f06724314a1e6ba89494e10647d9c277940a5da6/parser/ndt.go#L176)):*

I'm guessing this is a copy/paste bug. Please use "ndt" or "ndt.ParseAndInsert" here.
</blockquote></details>

Done.

---

*[parser/ndt.go, line 176 at r3](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lmPgav6pVCZWIz2bq:-L4mB4LtWeRgQc32Vnas:b-6hzh1w) ([raw file](https://github.com/m-lab/etl/blob/f06724314a1e6ba89494e10647d9c277940a5da6/parser/ndt.go#L176)):*

defer statements will run even if a panic has not occurred. But `recover()` will return `nil` if there is no panic.

Does this statement work as intended when `ParseAndInsert` returns a non-nil error without a panic? Or, is `err` reset to nil? (p.s. this is an example of reasoning about the system without panic re-propagation.)
</blockquote></details>

Ah - yikes.  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430)*

stephen-soltesz:
Review status: 1 of 7 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[cmd/etl_worker/etl_worker.go, line 107 at r4](https://reviewable.io:443/reviews/m-lab/etl/430#-L4mT13lrQWBcir5_O9u:-L4mT13lrQWBcir5_O9v:bt64gal) ([raw file](https://github.com/m-lab/etl/blob/cd5b57a068f3a35695e52583f2339c7323a31ec9/cmd/etl_worker/etl_worker.go#L107)):*
> ```Go
> 	// This will recover from any panics.
> 	defer func() {
> 		err := etl.CatchPanic(recover(), "worker")
> ```

Three parameters.

---

*[etl/globals.go, line 216 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lGEnovXwFP91bZ-WT:-L4mNKiyPOjrIa0OunXB:bdefm) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L216)):*

https://github.com/golang/go/wiki/PanicAndRecover asserts a convention of not allowing panics to cross package boundaries.  So the panics in Task and Parser should be captured and converted to errors.

Since etl_worker is in main, we could allow that panic to propagate and terminate that go routine.  However, I'd prefer to handle that error explicitly.

Good point about whether that error will result in the incoming task failing.  I've added explicit code to return an http error.  I believe that should provide equivalent behavior.

</blockquote></details>

Aaah.  "no *explicit* panic() should be allowed to cross a package boundary " -- emphasis added.

That actually makes sense to me. If a package calls `panic` explicitly, then it should also clean up the mess before passing control back to your caller. I read that practice statement as attempting to explicitly discourage methods from declaring a panic as valid return type as is common in Python, C++, and Java.

My worry comes from the accidental panics rather than explicit calls to `panic()`. Was the recent panic bug due to a call to `panic()` or was it accidental? Can `recover` tell the difference? If it can, then we might both be satisfied.

I still assert that attempting to recover from a "segfault" (or its equivalent) negatively changes how well we are able to reason about the system.

Will we still get stack traces in the logs to help debug recovered panic situations?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430)*

yachang:
Review status: 1 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[etl/globals.go, line 216 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lGEnovXwFP91bZ-WT:-L4nL0EK8-FOWvh_BcFO:b-mei1ke) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L216)):*

Aaah.  "no *explicit* panic() should be allowed to cross a package boundary " -- emphasis added.

That actually makes sense to me. If a package calls `panic` explicitly, then it should also clean up the mess before passing control back to your caller. I read that practice statement as attempting to explicitly discourage methods from declaring a panic as valid return type as is common in Python, C++, and Java.

My worry comes from the accidental panics rather than explicit calls to `panic()`. Was the recent panic bug due to a call to `panic()` or was it accidental? Can `recover` tell the difference? If it can, then we might both be satisfied.

I still assert that attempting to recover from a "segfault" (or its equivalent) negatively changes how well we are able to reason about the system.

Will we still get stack traces in the logs to help debug recovered panic situations?
</blockquote></details>

I am curious to see a test binary (with this PR, but revert my one-line fix with PT panic) deployed on sandbox and check how log changed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430)*

gfr10598:
Reverted panic check in NDT.  Now just has lint fixes.
Reverted panic check in Task.
Fixed CatchPanic so that it propagates error properly.

---

Review status: 1 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[etl/globals.go, line 216 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lGEnovXwFP91bZ-WT:-L4mYIos8NLV5lQDo4-p:b-pic53d) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L216)):*

I am curious to see a test binary (with this PR, but revert my one-line fix with PT panic) deployed on sandbox and check how log changed.
</blockquote></details>

Ah - good distinction.  I can work with this!

I've changed this function to PanicToErr, and added AddPanicMetrics, which just increments metric, writes a log, and repanics.

Removed the calls from Task and NDT parser, so now it is only top level.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430#-:-L4mXplPdsKgtiKDKYvd:bfr67gp)*

gfr10598:
Review status: 1 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[cmd/etl_worker/etl_worker.go, line 107 at r4](https://reviewable.io:443/reviews/m-lab/etl/430#-L4mT13lrQWBcir5_O9u:-L4qUztyJUi65fZpXJSy:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/cd5b57a068f3a35695e52583f2339c7323a31ec9/cmd/etl_worker/etl_worker.go#L107)):*

Three parameters.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430)*

stephen-soltesz:
:lgtm: - thank you for exploring more options!

---

Reviewed 1 of 7 files at r4, 2 of 5 files at r5.
Review status: 4 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[etl/globals.go, line 216 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lGEnovXwFP91bZ-WT:-L4qZFn_bZK2lOoDDL5f:b-a4uihw) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L216)):*

Ah - good distinction.  I can work with this!

I've changed this function to PanicToErr, and added AddPanicMetrics, which just increments metric, writes a log, and repanics.

Removed the calls from Task and NDT parser, so now it is only top level.

</blockquote></details>

This separation is very good.

Looking back, I see that what I was calling "panics" above is better referred to as "runtime errors". The panic mechanism is how runtime errors change control flow. But, because I've never interacted with panics for anything other than runtime errors, I conflated them in my mind.

It turns out that `runtime.Error` is the type that runtime errors send as a panic. So, these can be distinguished from other types at runtime.

I would prefer that PanicToErr use this distinction, even if just to print a scarier message. I'm still not sure under what circumstances I would prefer to use PanicToErr.

---

*[etl/globals.go, line 233 at r5](https://reviewable.io:443/reviews/m-lab/etl/430#-L4qZJsPgt-c4qTVYKUe:-L4qZJsQwMcpOPKD39W4:b7kci3k) ([raw file](https://github.com/m-lab/etl/blob/0a3c3ce9b53587545fa9a190169ad573b4f82fbf/etl/globals.go#L233)):*
> ```Go
> //        ...
> //    }
> func AddPanicMetric(r interface{}, tag string) {
> ```

`CountPanics`? `IncPanicCount`? This might be a good candidate for moving to `metrics/metrics.go`

---

*[etl/globals.go, line 247 at r5](https://reviewable.io:443/reviews/m-lab/etl/430#-L4q_A0wE8wdYVUreRgK:-L4q_A0wE8wdYVUreRgL:b-ajpuuy) ([raw file](https://github.com/m-lab/etl/blob/0a3c3ce9b53587545fa9a190169ad573b4f82fbf/etl/globals.go#L247)):*
> ```Go
> 
> // PanicToErr captures panics and converts them to
> // errors.  Use with care, as panic may mean that state
> ```

"extreme care".  And something like "continuing to execute after a runtime error may have undefined behavior."

---

*[etl/globals_test.go, line 171 at r5](https://reviewable.io:443/reviews/m-lab/etl/430#-L4qXiKfO-kiprEOUnUd:-L4qXiKfO-kiprEOUnUe:b-1g1pb2) ([raw file](https://github.com/m-lab/etl/blob/0a3c3ce9b53587545fa9a190169ad573b4f82fbf/etl/globals_test.go#L171)):*
> ```Go
> }
> 
> func RePanic() {
> ```

nit: please lower-case the first char of these helper functions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430#-:-L4qcL5fp-aOtgOV4zIb:bclmiha)*

gfr10598:
Review status: 4 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[etl/globals.go, line 216 at r2](https://reviewable.io:443/reviews/m-lab/etl/430#-L4lGEnovXwFP91bZ-WT:-L4qet1bZvBno2HD5emK:b-7yd1q4) ([raw file](https://github.com/m-lab/etl/blob/70ad9a137422921a09b648c351d9d08a86ce7d85/etl/globals.go#L216)):*

This separation is very good.

Looking back, I see that what I was calling "panics" above is better referred to as "runtime errors". The panic mechanism is how runtime errors change control flow. But, because I've never interacted with panics for anything other than runtime errors, I conflated them in my mind.

It turns out that `runtime.Error` is the type that runtime errors send as a panic. So, these can be distinguished from other types at runtime.

I would prefer that PanicToErr use this distinction, even if just to print a scarier message. I'm still not sure under what circumstances I would prefer to use PanicToErr.
</blockquote></details>

Added a TODO.

---

*[etl/globals.go, line 233 at r5](https://reviewable.io:443/reviews/m-lab/etl/430#-L4qZJsPgt-c4qTVYKUe:-L4qef6mGelj0C9q_5cr:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/0a3c3ce9b53587545fa9a190169ad573b4f82fbf/etl/globals.go#L233)):*

`CountPanics`? `IncPanicCount`? This might be a good candidate for moving to `metrics/metrics.go`
</blockquote></details>

Done.

---

*[etl/globals.go, line 247 at r5](https://reviewable.io:443/reviews/m-lab/etl/430#-L4q_A0wE8wdYVUreRgK:-L4qeeUn5m0Wpeqa0LLd:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/0a3c3ce9b53587545fa9a190169ad573b4f82fbf/etl/globals.go#L247)):*

"extreme care".  And something like "continuing to execute after a runtime error may have undefined behavior."
</blockquote></details>

Done.

---

*[etl/globals_test.go, line 171 at r5](https://reviewable.io:443/reviews/m-lab/etl/430#-L4qXiKfO-kiprEOUnUd:-L4qdnB21kiZPXmEO-lM:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/0a3c3ce9b53587545fa9a190169ad573b4f82fbf/etl/globals_test.go#L171)):*

nit: please lower-case the first char of these helper functions.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/430)*

PR #443, opened 2018-02-12
Add new metrics to disco parser, prepare for generic size metrics

stephen-soltesz:
This change enhances the etl metrics and adds new metrics to disco.

Specifically, the http status code text is now a label on HistogramDuration, making it possible to report QPS, successful request latency, and error rates.

As well, this change adds TODOs to generalize four metrics currently only used by NDT.

The four categories are:

* bytes / test
* rows / test
* bytes / row
* fields / row

Currently these metrics are NDT specific, so eventually their names should be changed, but not before establishing that they can be used by multiple parsers.

- FileSizeHistogram - bytes/test
- EntryFieldCountHistogram - rows/test
- RowSizeHistogram - bytes/row
- DeltaNumFieldsHistogram - fields/row

The first part of this is havign DISCO use them with smaller values added to EntryFieldCountHistogram and FileSizeHistogram.

A demonstration of these metrics in the prototype-combined parser dashboard (very bottom).
DISCO:
http://status.mlab-sandbox.measurementlab.net:3000/dashboard/db/pipeline-disco?orgId=1&var-interval=2m&var-table=switch&var-service=etl-disco-parser&from=now-1h&to=now&var-show_count=sum

NDT:
http://status.mlab-sandbox.measurementlab.net:3000/dashboard/db/pipeline-disco?orgId=1&var-interval=2m&var-table=ndt&var-service=etl-ndt-parser&from=now-1h&to=now&var-show_count=sum

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/443)

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -371,6 +371,8 @@ var (
 		[]string{"table", "phase", "retries", "status"},
 	)
 
+	// TODO(dev): bytes/row - generalize this metric for any file type.
+	//
 	// RowSizeHistogram provides a histogram of bq row json sizes.  It is intended primarily for
 	// NDT, so the bins are fairly large.  NDT average json is around 200K
 	//
@@ -395,6 +397,8 @@ var (
 		[]string{"table"},
 	)
 
+	// TODO(dev): fields/row - generalize this metric for any file type.
+	//
 	// DeltaNumFieldsHistogram provides a histogram of snapshot delta field counts.  It is intended primarily for
 	// NDT.  Typical is about 13, but max might be up to 120 or so.
 	//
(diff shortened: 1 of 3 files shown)


gfr10598:
:lgtm:
Still looks good to me.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/443#-:-L5B5ex1L2MoO0Q4f266:bc990af)*

gfr10598:
DeltaNumFieldsHistogram - fields/delta, for snapshots that are different from previous snapshot.
EntryFieldCountHistogram - fields/test, which is approx fields/delta * num_deltas.  In NDT this does not include the final snapshot, which reports all of the fields.  But when deltas are included, they overwhelm the final snapshot by a couple orders of magnitude IIRC.

So, for things other than NDT, I think the EntryFieldCount would be most sensibly replaced with just bytes/row.  This is non-trivial to compute, but easy to see the averages in bigquery.  Where we use a map, I wonder if there is an easy way to estimate the memory footprint of the map, which might be more useful.

stephen-soltesz:
Thank you for clarifying EntryFieldCountHistogram. This is still okay. Basically, I'm working to establish a simple nomenclature that we can use across parsers so that we can use as much as possible a common dashboard. As well, a shared nomenclature and dashboard gives a concrete target when creating new parsers. Normalizing these where possible will improve the quality and maintainability of these systems over time.

PR #444, opened 2018-02-12
Fix: Use explicit key path for ndt-batch

stephen-soltesz:
This change uses the explicit key path for ndt-batch. Evidently the previous form resulted in a file name like `/tmp/.json`. 

This kind of bug may be an argument for a stand-along deploy script.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/444)

PR #445, opened 2018-02-12
Revert "Fix: Use explicit key path for ndt-batch"

stephen-soltesz:
The fix in https://github.com/m-lab/etl/pull/444 accidentally combined the unreviewed change from https://github.com/m-lab/etl/pull/443.

This change reverts both so that they can be merged independently.

This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/445)


stephen-soltesz:
We're not doing this now.

PR #476, opened 2018-03-20
Combine StartTimeStamp with StartTimeUsec as int64 value

stephen-soltesz:
This change adds support to the Sidestream parser for combining the StartTimeStamp and StartTimeUsec web100 values into a single int64 value.

This change reports StartTimeStamp in the same way as it is reported by the NDT parser.

Unfortunately, the sidestream and ndt parsers use incompatible datatypes for holding the web100 values (schema.Web100Snap struct or sidestream, schema.Web100ValueMap for ndt), so there is no shared code between them for handling the conversion.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/476)

### DIFF
--- a/parser/ss.go
+++ b/parser/ss.go
@@ -181,10 +181,6 @@ func PopulateSnap(ss_value map[string]string) (schema.Web100Snap, error) {
 			if err != nil {
 				return *snap, err
 			}
-			if key == "StartTimeStamp" {
-				// Combine the StartTimeStamp and StartTimeUsec values.
-				value = value*1000000 + startTimeUsec
-			}
 			x.SetInt(value)
 		case "string":
 			x.Set(reflect.ValueOf(ss_value[key]))
@@ -198,6 +194,9 @@ func PopulateSnap(ss_value map[string]string) (schema.Web100Snap, error) {
 			}
 		}
 	}
+	// Combine the StartTimeStamp and StartTimeUsec values.
+	snap.StartTimeStamp = snap.StartTimeStamp*1000000 + startTimeUsec
+
 	// TODO: check whether snap has valid LocalAddress, RemAddress. Return error if not.
 	return *snap, nil
 }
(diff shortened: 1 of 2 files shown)


gfr10598:
:lgtm:
One suggestion.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[parser/ss.go, line 184 at r1](https://reviewable.io:443/reviews/m-lab/etl/476#-L84F8r5tt_k9nkNHmtu:-L84F8r5tt_k9nkNHmtv:b12teec) ([raw file](https://github.com/m-lab/etl/blob/00a6c729ccfd9a9fa38f3b469111f662b5f8cb7e/parser/ss.go#L184)):*
> ```Go
> 				return *snap, err
> 			}
> 			if key == "StartTimeStamp" {
> ```

How about doing this after the for loop completes?  I think that would read a little better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/476#-:-L84FQJOyMgM3aOJP6jp:b-2851ux)*

stephen-soltesz:
Thank you for the review!

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[parser/ss.go, line 184 at r1](https://reviewable.io:443/reviews/m-lab/etl/476#-L84F8r5tt_k9nkNHmtu:-L84FyhU6k81maNtXA3r:bcvjgzp) ([raw file](https://github.com/m-lab/etl/blob/00a6c729ccfd9a9fa38f3b469111f662b5f8cb7e/parser/ss.go#L184)):*

How about doing this after the for loop completes?  I think that would read a little better.
</blockquote></details>

Good idea. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/etl/476#-:-L84G24geyTk15ncKf6s:b-rrkdbt)*

stephen-soltesz:
Verified in mlab-sandbox that the new parser appears to report StartTimeStamp as a microsecond value.

PR #518, opened 2018-06-12
Compute project and dataset

gfr10598:
This PR:
 1. adds a Project parameter to InserterParams,
 2. regularizes selection of project, dataset, and table in etl/globals.go.
 3. eliminates export of previously exported maps in globals.go, (replacing with methods)
 4. switches from singleton BQ client to client per task, to allow sidestream tasks to use a different client associated with possibly different project.

Also implements TODO to cleanup obsolete dataset parameter.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/518)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -247,14 +247,14 @@ deploy:
 
 ######################################################################
 #  Prod deployments
-#  Deployed on manual tagging with prod-*, or (group)-prod-*
+#  Deployed on manual tagging with prod-*, ndt-prod-*, or small-prod-*
 #  Should be used AFTER code review, commit to integration, and staging soak.
-#  Triggers when *ANY* branch is tagged with prod-*'
+#  Triggers when *ANY* branch is tagged with one of these tags'
 ######################################################################
 
 ###################### SMALLER ETL SERVICES ###############################
 # Deploys smaller production services: PT, SS, DISCO
-# Triggers when *ANY* branch is tagged with qp-prod-* OR prod-*
+# Triggers when *ANY* branch is tagged with small-prod-* OR prod-*
 # NOTE: See later target for ndt-prod-*, which also triggers on prod-*, and
 #       deploys the NDT daily and batch pipelines.
 # NOTE: Failure in one of the deployments will terminate the deployment sequence,
@@ -280,7 +280,7 @@ deploy:
 
 ###################### NDT SERVICES ###############################
 # Deploys BOTH STREAMING AND BATCH services, along with queue config.
-# Triggers when *ANY* branch is tagged with qp-prod-* OR prod-*
+# Triggers when *ANY* branch is tagged with ndt-prod-* OR prod-*
 - provider: script
   script:
     $TRAVIS_BUILD_DIR/travis/activate_service_account.sh SERVICE_ACCOUNT_mlab_oti
--- a/bq/bq_test.go
+++ b/bq/bq_test.go
@@ -3,6 +3,7 @@ package bq_test
 import (
 	"errors"
 	"log"
+	"os"
 	"testing"
 	"time"
 
@@ -90,6 +91,31 @@ func TestBasicInsert(t *testing.T) {
 		t.Error("Should have accepted three rows")
 	}
 	in.Flush()
+}
+
+func TestInsertConfig(t *testing.T) {
+	os.Setenv("GCLOUD_PROJECT", "mlab-oti")
+	in, err := bq.NewInserter(etl.SS, time.Now())
+	if err != nil {
+		t.Fatal(err)
+	}
+	if in.Dataset() != "private" {
+		t.Errorf("Want private, got %s", in.Dataset())
+	}
+	if in.Project() != "mlab-oti" {
+		t.Errorf("Want private, got %s", in.Project())
+	}
+
+	in, err = bq.NewInserter(etl.NDT, time.Now())
+	if err != nil {
+		t.Fatal(err)
+	}
+	if in.Dataset() != "base_tables" {
+		t.Errorf("Want private, got %s", in.Dataset())
+	}
+	if in.Project() != "measurement-lab" {
+		t.Errorf("Want measurement-lab, got %s", in.Project())
+	}
 }
 
 func TestBufferingAndFlushing(t *testing.T) {
--- a/bq/bq_test.go
+++ b/bq/bq_test.go
@@ -93,7 +93,13 @@ func TestBasicInsert(t *testing.T) {
 	in.Flush()
 }
 
+// This does not currently work with mlab-testing credentials, and we
+// probably don't want it to.  So disabling it in travis.
 func TestInsertConfig(t *testing.T) {
+	_, isTravis := os.LookupEnv("TRAVIS")
+	if isTravis {
+		return
+	}
 	os.Setenv("GCLOUD_PROJECT", "mlab-oti")
 	in, err := bq.NewInserter(etl.SS, time.Now())
 	if err != nil {
@@ -103,7 +109,7 @@ func TestInsertConfig(t *testing.T) {
 		t.Errorf("Want private, got %s", in.Dataset())
 	}
 	if in.Project() != "mlab-oti" {
-		t.Errorf("Want private, got %s", in.Project())
+		t.Errorf("Want mlab-oti, got %s", in.Project())
 	}
 
 	in, err = bq.NewInserter(etl.NDT, time.Now())
--- a/bq/bq_test.go
+++ b/bq/bq_test.go
@@ -46,7 +46,7 @@ func xTestRealPartitionInsert(t *testing.T) {
 		Item{Name: tag + "_x1", Count: 12, Foobar: 44}}
 
 	in, err := bq.NewBQInserter(
-		etl.InserterParams{"mlab_sandbox", "test2", "_20160201", 10 * time.Second, 1, 0 * time.Second}, nil)
+		etl.InserterParams{"mlab-testing", "dataset", "test2", "_20160201", 10 * time.Second, 1, 0 * time.Second}, nil)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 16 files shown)


gfr10598:
FYI had to fix a test that doesn't work in travis.

---

Review status: 0 of 1 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/etl/518#-:-LEpIa2TAcKDOkYI9t1U:b1tnicx)*

stephen-soltesz:
An open question about how the bigquery package uses a context with timeout, and a general suggestion for not hard-coding project names.

:lgtm:

---

Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[bq/insert.go, line 97 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpK5ybE0DlRR-cHnn3:-LEpK5ybE0DlRR-cHnn4:b-1nuvj4) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/bq/insert.go#L97)):*
> ```Go
> //===============================================================================
> 
> // GetClient returns an appropriate bigquery client for datatype.
> ```

s/datatype/project/ ? Datatype is not a parameter here.

---

*[bq/insert.go, line 106 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpKqDACtDx64q6r7rY:-LEpKqDACtDx64q6r7rZ:b-nmu6v7) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/bq/insert.go#L106)):*
> ```Go
> 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
> 	defer cancel()
> 	return bigquery.NewClient(ctx, project)
> ```

This is a question I'm still unclear on -- and it may be b/c some libraries behave differently. How is the `ctx` timeout used by the bigquery (and underlying) package?

In this case, afaict, the bigquery package uses the `ctx` to create an http client, which it reuses for all future connections / operations. If that's true, then I think it is a bug to run `cancel` before that context is used.

If it only used the ctx for initially setting up credentials, then this would be fine. But, it's not clear from the code that its only using it for that.

My read of the docs for `context.WithTimeout` is that this is equivalent to `WithDeadline(parent, time.Now().Add(timeout))` which changes this to an absolute time internally.

I ran into a related bug in the `epoxy_client` where I called `cancel` prematurely and behavior was random -- sometimes it would cause failures sometimes it would appear that nothing was wrong.

I guess, I'm asking -- is this correct? Are you sure? And, if so, can you help me understand why?

---

*[etl/globals.go, line 253 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpOfHg1btrswncYenC:-LEpOfHg1btrswncYenD:bqwzm3w) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/etl/globals.go#L253)):*
> ```Go
> 	project := os.Getenv("GCLOUD_PROJECT")
> 	// For production, all datatypes except SS write to tables in measurement-lab.
> 	if project == "mlab-oti" && dt != SS {
> ```

So, thinking back to a comment from the other day -- I think part of the pattern here that makes me uneasy is the hard-coding of project names. For example, it will be impossible to run staging deployments where the parser is writing to a table in a separate project for testing without code changes. It's not necessary for this PR, but perhaps an issue, would be to make the `GCLOUD_PROJECT` separate from `TARGET_PROJECT` -- but allow both to be read from the environment.

---

*[etl/globals_test.go, line 207 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpPSDSDLzR7UdFu2yp:-LEpPSDSDLzR7UdFu2yq:b-o0zwe3) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/etl/globals_test.go#L207)):*
> ```Go
> 		dt       etl.DataType
> 		isBatch  bool
> 		gproject string // Shouldn't matter
> ```

If the project doesn't matter, does it need to be a test parameter? Please hard code a static string below.

---

*[etl/globals_test.go, line 231 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpQHnz1pjb_fNswsDO:-LEpQHnz1pjb_fNswsDP:b-euhc7) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/etl/globals_test.go#L231)):*
> ```Go
> 	tests := []struct {
> 		dt       etl.DataType
> 		isBatch  bool // Shouldn't matter
> ```

Please don't include any parameters that don't matter? Or run the test method twice with both values to establish that the result is always the same?

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/etl/518#-:-LEpQnz2CblYALEzfYU4:b-9zudg)*

gfr10598:
Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[bq/insert.go, line 97 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpK5ybE0DlRR-cHnn3:-LEpUxE80R3knR9GnvAO:bs7wumb) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/bq/insert.go#L97)):*

s/datatype/project/ ? Datatype is not a parameter here.
</blockquote></details>

Done.

---

*[bq/insert.go, line 106 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpKqDACtDx64q6r7rY:-LEpUigR0xu0OPJ9KxK8:b-y0zbu8) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/bq/insert.go#L106)):*

This is a question I'm still unclear on -- and it may be b/c some libraries behave differently. How is the `ctx` timeout used by the bigquery (and underlying) package?

In this case, afaict, the bigquery package uses the `ctx` to create an http client, which it reuses for all future connections / operations. If that's true, then I think it is a bug to run `cancel` before that context is used.

If it only used the ctx for initially setting up credentials, then this would be fine. But, it's not clear from the code that its only using it for that.

My read of the docs for `context.WithTimeout` is that this is equivalent to `WithDeadline(parent, time.Now().Add(timeout))` which changes this to an absolute time internally.

I ran into a related bug in the `epoxy_client` where I called `cancel` prematurely and behavior was random -- sometimes it would cause failures sometimes it would appear that nothing was wrong.

I guess, I'm asking -- is this correct? Are you sure? And, if so, can you help me understand why?
</blockquote></details>

Good catch.  I was using this for two purposes that really are very different.
I've hardcoded a more appropriate value here, and updated the field name, comments and value.
I'll watch this in staging to see that it functions properly.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/etl/518)*

stephen-soltesz:
Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[bq/insert.go, line 106 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpKqDACtDx64q6r7rY:-LEpWVQR2N5ZXf9GYRuC:byr565e) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/bq/insert.go#L106)):*

Good catch.  I was using this for two purposes that really are very different.
I've hardcoded a more appropriate value here, and updated the field name, comments and value.
I'll watch this in staging to see that it functions properly.

</blockquote></details>

I cannot verify that "[the context] is not used by the client" -- can we walk through the code together to verify?

In particular, it looks like the `bqs` (bigquery service) allocated in NewClient uses the ctx to create an http client -- https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/bigquery/bigquery.go#L79

The `bqs` variable is used throughout the bigquery package to access all other bq api functions.

To be clear -- I'm not saying this is incorrect usage -- I'm saying if it is correct it is unclear to me why.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/etl/518)*

gfr10598:
Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[bq/insert.go, line 106 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpKqDACtDx64q6r7rY:-LEppnenEtON0YNt8MZz:b-7p74or) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/bq/insert.go#L106)):*

I cannot verify that "[the context] is not used by the client" -- can we walk through the code together to verify?

In particular, it looks like the `bqs` (bigquery service) allocated in NewClient uses the ctx to create an http client -- https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/bigquery/bigquery.go#L79

The `bqs` variable is used throughout the bigquery package to access all other bq api functions.

To be clear -- I'm not saying this is incorrect usage -- I'm saying if it is correct it is unclear to me why.
</blockquote></details>

I've tweaked the code a bit, because I think it was misleading.  Added a comment.  Let's look at it together.

---

*[etl/globals.go, line 253 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpOfHg1btrswncYenC:-LEpmLRp1N8-9A7hggNr:b-h3l3sk) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/etl/globals.go#L253)):*

So, thinking back to a comment from the other day -- I think part of the pattern here that makes me uneasy is the hard-coding of project names. For example, it will be impossible to run staging deployments where the parser is writing to a table in a separate project for testing without code changes. It's not necessary for this PR, but perhaps an issue, would be to make the `GCLOUD_PROJECT` separate from `TARGET_PROJECT` -- but allow both to be read from the environment.
</blockquote></details>

We already have BIGQUERY_PROJECT, which this code is obsoleting.  I could keep it, and have it override the computed value if provided.
I'll code that up and see what you think.

---

*[etl/globals_test.go, line 207 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpPSDSDLzR7UdFu2yp:-LEpmFrl0_0wB73Lr_4R:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/etl/globals_test.go#L207)):*

If the project doesn't matter, does it need to be a test parameter? Please hard code a static string below.
</blockquote></details>

Done.

---

*[etl/globals_test.go, line 231 at r2](https://reviewable.io/reviews/m-lab/etl/518#-LEpQHnz1pjb_fNswsDO:-LEpmEU2-bAR1h_15bHh:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/2fe67a69f8111c7ede8fb63547a285e97fdd1d97/etl/globals_test.go#L231)):*

Please don't include any parameters that don't matter? Or run the test method twice with both values to establish that the result is always the same?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/etl/518)*

PR #523, opened 2018-06-14
Ndt sandbox async flush cleanup

gfr10598:
This PR:
1. Adapts BQInserter.Flush to FlushSlice, and adds support for asynchronous flushing.
2. Adds FlushAsync() which executes the BQ flush in a goroutine.
3. Add BQInserter.AddRow(), which adds a row without flushing.
4. Updates SS parser to use AddRow(), and FlushAsync()
5. Updates and improves unit tests.
6. Tweaks organization of BQInserter fields.
7. Tweaks parameters in test case BQInserters.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/523)

### DIFF
--- a/bq/bq_test.go
+++ b/bq/bq_test.go
@@ -34,6 +34,16 @@ type Item struct {
 	Foobar int `json:"foobar"`
 }
 
+func standardInsertParams(bufferSize int) etl.InserterParams {
+	return etl.InserterParams{
+		Project: "mlab-testing", Dataset: "dataset", Table: "table",
+		Suffix:        "",
+		BufferSize:    bufferSize,
+		PutTimeout:    10 * time.Second,
+		MaxRetryDelay: 1 * time.Second,
+	}
+}
+
 //==================================================================================
 // These tests hit the backend, to verify expected behavior of table creation and
 // access to partitions.  They deliberately have a leading "x" to prevent running
@@ -46,8 +56,9 @@ func xTestRealPartitionInsert(t *testing.T) {
 		Item{Name: tag + "_x0", Count: 17, Foobar: 44},
 		Item{Name: tag + "_x1", Count: 12, Foobar: 44}}
 
-	in, err := bq.NewBQInserter(
-		etl.InserterParams{"mlab-testing", "dataset", "test2", "_20160201", 1, 10 * time.Second, 1 * time.Second}, nil)
+	params := standardInsertParams(1)
+	params.Suffix = "_20160201"
+	in, err := bq.NewBQInserter(params, nil)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -73,9 +84,8 @@ func TestBasicInsert(t *testing.T) {
 		Item{Name: tag + "_x0", Count: 17, Foobar: 44},
 		Item{Name: tag + "_x1", Count: 12, Foobar: 44}}
 
-	in, err := bq.NewBQInserter(
-		etl.InserterParams{"mlab-testing", "dataset", "test2", "", 1, 10 * time.Second, 1 * time.Second},
-		fake.NewFakeUploader())
+	params := standardInsertParams(1)
+	in, err := bq.NewBQInserter(params, fake.NewFakeUploader())
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -131,9 +141,7 @@ func TestBufferingAndFlushing(t *testing.T) {
 
 	// Set up an Inserter with a fake Uploader backend for testing.
 	// Buffer 3 rows, so that we can test the buffering.
-	in, err := bq.NewBQInserter(
-		etl.InserterParams{"mlab-testing", "dataset", "test2", "", 3, 10 * time.Second, 1 * time.Second},
-		fake.NewFakeUploader())
+	in, err := bq.NewBQInserter(standardInsertParams(3), fake.NewFakeUploader())
 	if err != nil {
 		log.Printf("%v\n", err)
 		t.Fatal()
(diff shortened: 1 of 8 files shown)


gfr10598:
PTAL

---

Review status: 0 of 1 LGTMs obtained

---

*[bq/bq_test.go, line 199 at r2](https://reviewable.io/reviews/m-lab/etl/523#-LF--qEt0SU_YGJi3v1K:-LF-ph5T-IcT_6YEILiL:b3lxrh9) ([raw file](https://github.com/m-lab/etl/blob/230a720003509a2825be3da35e34fac26e354863/bq/bq_test.go#L199)):*

This struct is getting complex enough that the absence of field names obscures the readability. I looked at the version of `InserterParams` at HEAD to figure out what which field was the BufferSize. And, then discovered that this PR changes the order. Explicit field name declarations make this more resilient to struct changes and easier to interpret.
</blockquote></details>

Ok.  I'm actually conflicted about named params, because it allows you to accidentally default some fields, which has caused bugs for me several times, particularly after adding a field.  I'll update these, though.  Actually, I'll use a small helper function, too.

---

*[bq/insert.go, line 48 at r2](https://reviewable.io/reviews/m-lab/etl/523#-LF-0pCN2c6Yg-JtgvZc:-LF-q5jQ4UgypThrF9DC:blwrbee) ([raw file](https://github.com/m-lab/etl/blob/230a720003509a2825be3da35e34fac26e354863/bq/insert.go#L48)):*

It's fine if so, but just checking -- do these need to be public?
</blockquote></details>

Nope.  Just a bad habit of mine, left over from using all caps for constants in C++.

---

*[bq/insert.go, line 153 at r2](https://reviewable.io/reviews/m-lab/etl/523#-LF-79XjBuwFKlNbviCF:-LF-sWRG2exfxhSw8sY8:bido9gh) ([raw file](https://github.com/m-lab/etl/blob/230a720003509a2825be3da35e34fac26e354863/bq/insert.go#L153)):*

I don't understand why mutex is unsuitable for this use case. What am I missing?
</blockquote></details>

Hmm.  I'm accustomed to mutex restrictions (or maybe conventions) requiring the same thread to lock and unlock the mutex.  That is perhaps not true in golang, but it still seems natural to me, so I'm reluctant to violate that.

Actually, this discussion is interesting: https://github.com/golang/go/issues/9201

---

*[bq/insert.go, line 230 at r2](https://reviewable.io/reviews/m-lab/etl/523#-LF-2bL096ei25sB_WOB:-LF-tDZ0CH8ptAKnTH-o:blxifeu) ([raw file](https://github.com/m-lab/etl/blob/230a720003509a2825be3da35e34fac26e354863/bq/insert.go#L230)):*

Why add `tableBase` as a parameter? Is this approach used consistently within this file?
</blockquote></details>

Oops.  That was required in an earlier draft, but is no longer.  I'll restore it.

---

*[etl/etl.go, line 35 at r2](https://reviewable.io/reviews/m-lab/etl/523#-LF-534Y2gzdd08jWEz-:-LF-u4F9-QXK3rxe1Cyq:b-2syeng) ([raw file](https://github.com/m-lab/etl/blob/230a720003509a2825be3da35e34fac26e354863/etl/etl.go#L35)):*

Deprecated in favor of what?
</blockquote></details>

Added comment.  Thanks for pointing out.

---

*[parser/ss.go, line 272 at r2](https://reviewable.io/reviews/m-lab/etl/523#-LF-5kFd3u9fN6d13-Is:-LF-u9jr3vXUl46Tq1DV:b-redv8v) ([raw file](https://github.com/m-lab/etl/blob/230a720003509a2825be3da35e34fac26e354863/parser/ss.go#L272)):*

Why not `AddRow`?
</blockquote></details>

Because I missed this.  8-(  Too bad the deprecation warnings aren't more prominent.  Actually, I'm not even sure what tool produces them.

Hmm - a new tool:
go get -u honnef.co/go/tools/cmd/staticcheck
staticcheck ./...
go get -u honnef.co/go/tools/cmd/megacheck
megacheck ./...

It doesn't warn about deprecated tags, but does generate some other useful warning we should probably clean up.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/etl/523#-:-LF-wN6t5l506beYrjU9:bb71pro)*

stephen-soltesz:
:lgtm:

---

Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[bq/bq_test.go, line 199 at r2](https://reviewable.io/reviews/m-lab/etl/523#-LF--qEt0SU_YGJi3v1K:-LF2oo3b1KcSlTKkYN-f:b4lqls4) ([raw file](https://github.com/m-lab/etl/blob/230a720003509a2825be3da35e34fac26e354863/bq/bq_test.go#L199)):*

Ok.  I'm actually conflicted about named params, because it allows you to accidentally default some fields, which has caused bugs for me several times, particularly after adding a field.  I'll update these, though.  Actually, I'll use a small helper function, too.
</blockquote></details>

In this case, I would strive for the following balance.

My understanding is that Go encourages having meaningful zero values. If zero values have caused bugs in the past, I would look to the logic that depends on structs and consider how zero values will be interpreted when first creating the code. I know that's easier to say than do consistently. As well, when structs are complex enough to require special handling to prevent zero values, maybe by assigning default values, in this case defining a `New*` function can solve that case. Which looks like what you've done. :+1: Thank you.

---

*[bq/insert.go, line 153 at r2](https://reviewable.io/reviews/m-lab/etl/523#-LF-79XjBuwFKlNbviCF:-LF2qXyo2Kg-qEN0iSVk:bu64ix3) ([raw file](https://github.com/m-lab/etl/blob/230a720003509a2825be3da35e34fac26e354863/bq/insert.go#L153)):*

Hmm.  I'm accustomed to mutex restrictions (or maybe conventions) requiring the same thread to lock and unlock the mutex.  That is perhaps not true in golang, but it still seems natural to me, so I'm reluctant to violate that.

Actually, this discussion is interesting: https://github.com/golang/go/issues/9201

</blockquote></details>

Excellent. Carry on.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/etl/523#-:-LF2tDwKAH64rFWQ4Np6:bnfp4nl)*

gfr10598:
Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[bq/insert.go, line 153 at r2](https://reviewable.io/reviews/m-lab/etl/523#-LF-79XjBuwFKlNbviCF:-LF30-tLDdZhJJmf6igj:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/230a720003509a2825be3da35e34fac26e354863/bq/insert.go#L153)):*

Excellent. Carry on.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/m-lab/etl/523)*

PR #579, opened 2018-10-19
Rollback instance & rate increases to unblock k8s platform work while…

stephen-soltesz:
DO NOT SUBMIT.

This PR reverts the increase instance and rates for the batch pipeline because the excess resource utilization was preventing development work on the k8s platform while we wait for quota increases for IP addresses.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/579)


stephen-soltesz:
Obsolete.

PR #596, opened 2018-12-12
Use annotator structs

gfr10598:
Now that json tags are respected, we can switch to the annotator/api structs.

This also removes the EnableAnnotation, since we haven't used it in over a year.
(HMM - need to check with Stephen about disco, which has annotation disabled.)

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/596)

### DIFF
--- a/annotation/geo.go
+++ b/annotation/geo.go
@@ -12,72 +12,12 @@ import (
 	"strconv"
 	"time"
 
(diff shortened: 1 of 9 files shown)


gfr10598:
Checked the code.  Disco wasn't checking the env var anyway.

PR #599, opened 2018-12-13
Use the new annotator-service/api.GetAnnotations

gfr10598:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/599)

### DIFF
--- a/annotation/export_test.go
+++ b/annotation/export_test.go
@@ -1,0 +1,5 @@
+package annotation
+
+var (
+	BatchParseJSONGeoDataResponse = batchParseJSONGeoDataResponse
+)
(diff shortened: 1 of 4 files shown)


gfr10598:
In sandbox this causes complete insert failure (for etl-batch-parser).  Trying redeploying master just to batch using batch-sandbox-master branch.

gfr10598:
Hmm.   Looks like the problem is with SS and TraceRoute.
https://grafana.mlab-sandbox.measurementlab.net/d/000000037/pipeline-parser?orgId=1&var-datasource=Prometheus%20(mlab-sandbox)&var-interval=1h&var-table=All&var-service=All&var-show_count=sum%20by(service)&from=1544728394863&to=1544739194863

PR #600, opened 2019-01-02
Use the new annotation-service/api.GetAnnotations

gfr10598:
I've created a GetAnnotations function in annotation-service/api/v2 that wraps the http request and retries on ServiceUnavailable failures.  This keeps more of the repo specific code in the same annotation-service repo, though it does require a recompile of ETL to get new changes.

One question this raises - should we use dep to vendor the specific version of annotation-service used?  This is demonstrated in the batch-sandbox-retry-v2 branch.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/600)

### DIFF
--- a/annotation/export_test.go
+++ b/annotation/export_test.go
@@ -1,0 +1,5 @@
+package annotation
+
+var (
+	BatchParseJSONGeoDataResponse = batchParseJSONGeoDataResponse
+)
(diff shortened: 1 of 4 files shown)


pboothe:
Vendoring has always caused us headaches in the past - breaking library changes turned out to be far less frequent than us forgetting to update the vendoring, which made updating the vendoring painful, which meant we didn't do it.  I would prefer to live at head for any service where we own both ends of the service.

PR #661, opened 2019-05-22
Ndt sandbox base parser

gfr10598:
This migrates the NDT parser to use the parser.Base, and the new Annotatable interface.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/661)


gfr10598:
___
*[parser/ndt_test.go, line 120 at r2](https://reviewable.io/reviews/m-lab/etl/661#-Lf_JVOA67nTzg1Tahqz:-Lf_fzUU7L7Pd7U7GkYH:b-74sx03) ([raw file](https://github.com/m-lab/etl/blob/18eacb7ff97884b61376d86b587bff00a3768e3c/parser/ndt_test.go#L120)):*

> ```
> 	n.Flush()
> 	if ins.Accepted() != 2 {
> 		t.Fatalf("Failed to insert snaplog data.")
> 	}
> ```
 reason for deletion?
</blockquote></details>

It isn't deleted - just moved.   Previously there was parse,flush,parse,flush.  Now there is parse,parse,flush, so that there are two tests in the buffer for batch annotation.

PR #681, opened 2019-06-12
remove reflect.Kind().Elem()

gfr10598:
This is causing occasional panics from disco parser.  This is the quickest fix, though it will make debugging a little harder.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/681)


gfr10598:
Obsolete - fixed in another PR.

PR #711, opened 2019-07-06
fix ndt5 dependency

gfr10598:
change from legacy to ndt5 broke etl.
This fixes etl, so that it can build at head.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/711)


gfr10598:
Submitting TBR since currently broken at head.

PR #714, opened 2019-07-19
Update ETL parser to read ndt5 Result

stephen-soltesz:
This change removes support for the ndt "legacy" format for ndt5 results from the Go ndt-server. Instead, a new unified result schema records both ndt5 and ndt7 results. The current ETL worker will only recognize ndt5 results now due to the different directory hierarchy for ndt7 results. As well, this is the first time we've had two datatypes mapped to the same table, so gardener does not yet support ndt7 either. However, from 2019-07-19 forward, all results should use unified schemas.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/714)

### DIFF
--- a/cmd/update-schema/update.go
+++ b/cmd/update-schema/update.go
@@ -43,6 +43,13 @@ func CreateOrUpdatePT(project string, dataset string, table string) error {
 	return CreateOrUpdate(schema, project, dataset, table)
 }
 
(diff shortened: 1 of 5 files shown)


stephen-soltesz:
Part of https://github.com/m-lab/ndt-server/issues/143

Issue #719, opened 2019-08-19
Don't add mirror structs


pboothe:
Don't add mirror structs. Instead, make the schema generator better.

Eventually, the schema generator will need to be smart enough to know to generate fields for dead data.  I suggest that we adopt a solution that looks like:

To drop the bad1 and bad2 fields from the struct below, change from:
```go
type Data struct {
  id string
  bad1, bad2 [1000000]byte
}
```
to
```go
type Data struct {
  id string
  bad1 struct{} `schematype:[1000000]byte`
  bad2 struct{} `schematype:[1000000]byte`
}
```
That way the name stays there, but it is changed to being a zero-size data element.  The schema parser will need to be upgraded to support the relevant annotation.

(this came out of a discussion with @stephen-soltesz , but all flames should be directed at me and not him, as he didn't actually care that much about this issue, while I care a lot)

stephen-soltesz:
I think this is a more fundamental problem. There are three operations that we need to work:

1. Loading archival data into a struct -- i.e. parsing serialized JSON into a struct with all required fields
2. Inferring BigQuery Schema from a struct
3. Load data into bigquery from struct

In the case of JSON data that encodes a `map[string]string` we need an "Unmarshal" target struct that includes a `map[string]string` type. While `bigquery.InferSchema` normally fails for map types, we can add a field tag that tells bigquery to ignore that field when inferring the schema. We can also "fix" the schema return by `bigquery.InferSchema` so that it include equivalent fields for the map data as an array, e.g. `[]NameValue`. This could address the first two operations.

However, the third operation needs a source struct for loading data into BigQuery that has the same fields as the schema.

As proposed for the NDTResult struct, we would need to do one of the following:

1. Create full mirror structs to capture the two metadata field types, `map[string]string` for JSON, `[]NameValue` for BigQuery.
2. Create "mirror fields" in the native struct, one used only for JSON, the other used only for bigquery, where the parser copies data from one to the other.
  For example:
  ```
  type ArchivalData struct {
    ...
    ClientMetadataJSON map[string]string `bigquery:"-" json:"ClientMetadata"`
    ClientMetadataBQ []Name `bigquery:"ClientMetadata" json:"-"`
  }
  ```
3. Create the native struct so that it is compatible with bigquery.InferSchema from the beginning. i.e. no `map` types allowed.
[...]

Issue #697, opened 2019-06-18
NDT5 Parser -- Remove or add unit tests


stephen-soltesz:
After 2019-08-01 this parser should be removed or unit tests added

gfr10598:
m-lab/dev-tracker#402

stephen-soltesz:
Related to https://github.com/m-lab/etl/issues/719

PR #720, opened 2019-08-21
Update NDT Result Schema to include ClientMetadata

stephen-soltesz:
This is part of https://github.com/m-lab/etl/issues/697
This change depends on https://github.com/m-lab/go/pull/61

This change adds support for ClientMetadata in the NDT result object used by NDT5 & NDT7.

This pattern respects https://github.com/m-lab/etl/issues/719 by appending new ClientMetadata fields that are bigquery compatible to the schema inferred from the native NDT Result struct.

Also, rather than loading the result object into a Go struct, this change loads the result object into a `map[string]interface{}` type. This allows the restructuring of the ClientMetadata maps into records at run time.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/720)


stephen-soltesz:
Verified that this WAI with data collected from sandbox recorded with the new NDT result format.

Unified schema results without the ClientMetadata field in older data will continue to parse successfully. Any unified schema results *with* the ClientMetadata field will fail to parse.

PR #738, opened 2019-10-01
fix bug that caused runtime error

yachang:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/738)


parser/pt_test.go
```diff
@@ -57,6 +56,17 @@ func TestParseJsonSimple(t *testing.T) {
 	}
 }
 
+func TestParseJsonNoLinks(t *testing.T) {
```
pboothe: Please add a comment before the test block:
```
// Last object on the "type":"tracelb" line has "linkc":1 but no "links" set.
```

PR #743, opened 2019-10-08
Fix panic

yachang:
* fix the out of range panic when there is empty server IP.

* more

* unittest

* more

* minor

* more fix

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/743)


yachang:
dup of #741

pboothe:
Verified: This PR is exactly the same as #741, which has already been approved.  Stamping this one approved, and @yachang will close out the other.

PR #780, opened 2019-12-19
Code to poll gardener for work to do.

gfr10598:
This adds a polling loop that gets jobs from gardener, and processes all the corresponding files.  It is basically working in sandbox, but may need a little more refinement.

Won't currently work in staging (and isn't deployed to prod), for two reasons:
  1. archive-mlab-sandbox is hardcoded (because archive-measurement-lab is apparently not accessible from sandbox instances)
  2. gardener is currently in the wrong region, so private network is failing.

These will be fixed in upcoming PRs.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/780)

### DIFF
--- a/active/active_test.go
+++ b/active/active_test.go
@@ -5,12 +5,14 @@ package active_test
 import (
 	"context"
 	"log"
+	"net/http/httptest"
 	"os"
 	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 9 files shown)


gfr10598:
Replaced with other PRs.

PR #842, opened 2020-03-09
Add public status page

gfr10598:
This adds a public status/debug page to the k8s parser deployment.  Similar to etl-gardener.
Very useful for profiling and debugging.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/842)


critzo:
Will this public status page be suitable to include on the website?

PR #863, opened 2020-04-13
Factories for sink, source, and annotator

gfr10598:
This introduces factories that can be provided as the context for Job/Task processing.  The factories are used to create the TestSource, row.Sink, and Annotator.

This allows configuration at a high level for testing, or for using storage vs bigquery backend.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/863)


gfr10598:
obsolete

PR #963, opened 2021-01-25
Remove client and server annotations from ndt5 schema

stephen-soltesz:
The ndt7 schema does not include the server and client annotation fields because these fields are defined by the materialized join managed by etl-gardener.

This change removes the empty fields from the ndt5 schema because this data type should be managed like ndt7 wrt annotations.

Annotations pre-2020-03 will be exported by the annotation service to the annotation tables as part of future work.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/963)

### DIFF
--- a/schema/ndt5_result.go
+++ b/schema/ndt5_result.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/m-lab/go/cloud/bqx"
 	"github.com/m-lab/ndt-server/data"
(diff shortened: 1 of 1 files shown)


nkinkade (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/963)* status: :shipit: complete! 1 of 1 approvals obtained

PR #972, opened 2021-03-16
Prepare for cloudbuild migration 1 of 2

stephen-soltesz:
This change prepares for migration of the ETL build and deployment to cloudbuild, the first of two changes.

This change removes use of `os.Getenv` and `os.LookupEnv` throughout the sources in favor of flags that can be set using environment variables. In particular, this makes the `GitCommit` and `Version` variables set at build time, instead of reading them from the environment. In several other cases, this change means using global package variables until code is improved to set a "config" struct to pass around for access to these runtime flags.

NOTE: once this change is merged, the 'Version' will not be set correctly in mlab-staging for the universal gardener by the existing cloudbuild of the Dockerfile. The second change that completes the CB migration is ready and will resolve this.

TESTING: with the exception of the Version, I manually checked that current sandbox deployments appear to continue to work as before based on the [Pipeline][pipeline] grafana dashboards

[pipeline]: https://grafana.mlab-sandbox.measurementlab.net/d/000000037/pipeline-parser?orgId=1&var-datasource=Prometheus%20(mlab-sandbox)&var-interval=2m&var-table=ndt&var-service=All&var-show_count=sum%20by(service)&from=now-7d&to=now

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/972)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,7 +107,9 @@ script:
 
 # Clean build and prepare for deployment
 - cd $TRAVIS_BUILD_DIR/cmd/etl_worker &&
-  go build -v -ldflags "-X github.com/m-lab/go/prometheusx.GitShortCommit=$(git log -1 --format=%h)" .
+  go build -v -ldflags "-X github.com/m-lab/go/prometheusx.GitShortCommit=$(git log -1 --format=%h)
+    -X github.com/m-lab/etl/etl.Version=$TRAVIS_BRANCH$TRAVIS_TAG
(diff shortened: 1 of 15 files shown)


robertodauria (approved):
:lgtm:

> Reviewed 15 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/972#-:-MW-Iy0K-rFcjrYK2ql4:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #982, opened 2021-04-07
Add support for retired sites' annotations

robertodauria:
This PR adds support for retired sites' annotations to the `site` package. Annotations for retired sites are merged with the "current" annotations.json at loading time.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/982)


stephen-soltesz (changes requested):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/982)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[site/site.go, line 69 at r1](https://reviewable.io/reviews/m-lab/etl/982#-MXiByPI-KsK2ARCA6HN:-MXiByPI-KsK2ARCA6HO:b-mvnpyb) ([raw file](https://github.com/m-lab/etl/blob/5d26938134a639c9726bfc9945f9e36e8c36e675/site/site.go#L69)):*
> ```Go
> 	// Retry for up to 30 seconds.
> 	start = time.Now()
> 	for ; time.Since(start) < timeout; time.Sleep(time.Second) {
> ```

I see that this code mirrors the loop above -- but I think the timeout handling is overcomplicated.

Would it be equally correct to define a `context.WithTimeout` above and simply loop until success or context timeout? (imo, one context timeout should be enough for both fetches).

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/982)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[site/site.go, line 69 at r1](https://reviewable.io/reviews/m-lab/etl/982#-MXiByPI-KsK2ARCA6HN:-MY5oIje-XGkESvGtbyT:b-bj3u28) ([raw file](https://github.com/m-lab/etl/blob/5d26938134a639c9726bfc9945f9e36e8c36e675/site/site.go#L69)):*

I see that this code mirrors the loop above -- but I think the timeout handling is overcomplicated.

Would it be equally correct to define a `context.WithTimeout` above and simply loop until success or context timeout? (imo, one context timeout should be enough for both fetches).
</blockquote></details>

Done. As said earlier, it seems no connection is made when calling FromURL thus I removed the two retry loops and changed the one loop actually opening a connection to use `context.WithTimeout()`. Thoughts?

stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/982)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[site/site.go, line 69 at r1](https://reviewable.io/reviews/m-lab/etl/982#-MXiByPI-KsK2ARCA6HN:-MY6AiLxBZFoCWMbs1bX:b-vw8g67) ([raw file](https://github.com/m-lab/etl/blob/5d26938134a639c9726bfc9945f9e36e8c36e675/site/site.go#L69)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

Done. As said earlier, it seems no connection is made when calling FromURL thus I removed the two retry loops and changed the one loop actually opening a connection to use `context.WithTimeout()`. Thoughts?
</blockquote></details>

Please double check that the `ctx` used when creating the Google Storage Client is not saved/reused by that client -- otherwise after 30sec the next refresh would fail.
___
*[site/site.go, line 52 at r2](https://reviewable.io/reviews/m-lab/etl/982#-MY6BILeDlIpm7yZ5h4S:-MY6BILeDlIpm7yZ5h4T:b-1j9ooc) ([raw file](https://github.com/m-lab/etl/blob/aebe800206703a90e5722557207a061356e7625c/site/site.go#L52)):*
> ```Go
> // MustLoad loads the site annotation source.  Will try at least once,
> // and retry for up to timeout
> func MustLoad(timeout time.Duration) {
> ```

Will there need to be `Reload` logic? Having it would mean having one less thing to keep track of for the running parsers in production.
___
*[site/site.go, line 53 at r2](https://reviewable.io/reviews/m-lab/etl/982#-MY69xY2D95GJWDsQDn1:-MY69xY36xDVKLZgyjCH:b95iysg) ([raw file](https://github.com/m-lab/etl/blob/aebe800206703a90e5722557207a061356e7625c/site/site.go#L53)):*
> ```Go
> // and retry for up to timeout
> func MustLoad(timeout time.Duration) {
> 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
> ```

One second is probably not long enough. I suggest using the given `timeout` parameter here.

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/982)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[site/site.go, line 69 at r1](https://reviewable.io/reviews/m-lab/etl/982#-MXiByPI-KsK2ARCA6HN:-MY9rtQ10OyKWZ7w546g:b-fvmbhk) ([raw file](https://github.com/m-lab/etl/blob/5d26938134a639c9726bfc9945f9e36e8c36e675/site/site.go#L69)):*

Please double check that the `ctx` used when creating the Google Storage Client is not saved/reused by that client -- otherwise after 30sec the next refresh would fail.
</blockquote></details>

Hmm. Now that I dig further into the Provider implementations...
- httpsProvider ignores the `context` passed in `FromURL` and sets a default timeout of `time.Minute` for subsequent calls to `httpsProvider.Get(ctx)`.
- gcsProvider uses the `context` passed in `FromURL` to initialize a client with `storage.NewClient(ctx)`. Calls to `gcsProvider.Get(ctx)` use a separate context. It's still not clear to me whether the  context passed to `storage.NewClient(ctx)` is reused somehow, since Get(ctx) uses the passed-in context. Each call to `FromURL` creates a  new storage.Client -- it isn't reused.

However, in both cases I think the right context to pass to `FromURL` is `context.Background()`, while the `context.WithTimeout` is only used by the retry loop. I will also note in a comment that the timeout for an httpProvider must > 1 minute for the retry loop to make sense, since the HTTP call's timeout is always 1 minute and not configurable.
___
*[site/site.go, line 52 at r2](https://reviewable.io/reviews/m-lab/etl/982#-MY6BILeDlIpm7yZ5h4S:-MY9z0gW4ylyXOZzrBxL:b-o93jsw) ([raw file](https://github.com/m-lab/etl/blob/aebe800206703a90e5722557207a061356e7625c/site/site.go#L52)):*

Will there need to be `Reload` logic? Having it would mean having one less thing to keep track of for the running parsers in production.
</blockquote></details>

This package will likely be used by annotation-service's client rather than ETL. Yes, it will need some `Reload` logic. I've moved the loading logic to `Load` and turned `MustLoad` into a wrapper enforcing that there are no errors. I think annotation-service will need both, one at startup (failing if the annotations can't be loaded at all) and one for periodically reloading.
___
*[site/site.go, line 53 at r2](https://reviewable.io/reviews/m-lab/etl/982#-MY69xY2D95GJWDsQDn1:-MY9rsJBD-Ic204lAIGN:bbi32in) ([raw file](https://github.com/m-lab/etl/blob/aebe800206703a90e5722557207a061356e7625c/site/site.go#L53)):*

One second is probably not long enough. I suggest using the given `timeout` parameter here.
</blockquote></details>

Done. It was definitely a mistake.

stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/982)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[site/site.go, line 69 at r1](https://reviewable.io/reviews/m-lab/etl/982#-MXiByPI-KsK2ARCA6HN:-MYA_gm5AljzM8B14egD:bur4y2z) ([raw file](https://github.com/m-lab/etl/blob/5d26938134a639c9726bfc9945f9e36e8c36e675/site/site.go#L69)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

Hmm. Now that I dig further into the Provider implementations...
- httpsProvider ignores the `context` passed in `FromURL` and sets a default timeout of `time.Minute` for subsequent calls to `httpsProvider.Get(ctx)`.
- gcsProvider uses the `context` passed in `FromURL` to initialize a client with `storage.NewClient(ctx)`. Calls to `gcsProvider.Get(ctx)` use a separate context. It's still not clear to me whether the  context passed to `storage.NewClient(ctx)` is reused somehow, since Get(ctx) uses the passed-in context. Each call to `FromURL` creates a  new storage.Client -- it isn't reused.

However, in both cases I think the right context to pass to `FromURL` is `context.Background()`, while the `context.WithTimeout` is only used by the retry loop. I will also note in a comment that the timeout for an httpProvider must > 1 minute for the retry loop to make sense, since the HTTP call's timeout is always 1 minute and not configurable.
</blockquote></details>

Thank you for double checking. I feel like the storage package usage of the context to create the client is confusing.
___
*[site/site.go, line 54 at r3](https://reviewable.io/reviews/m-lab/etl/982#-MYAZ_Sr3sbJfHvfejGD:-MYAZ_Sr3sbJfHvfejGE:b-buzdwz) ([raw file](https://github.com/m-lab/etl/blob/0e06744425e7234f9d1a37f84d42275767bd06f4/site/site.go#L54)):*
> ```Go
> func MustLoad(timeout time.Duration) {
> 	err := Load(timeout)
> 	rtx.Must(err, "Could not load annotation db")
> ```

Please clarify "server annotations". Including the intended URLs might be helpful too. Depending on the error, the err string might not be enough.
___
*[site/site.go, line 59 at r3](https://reviewable.io/reviews/m-lab/etl/982#-MYAa9K8AZ2uOyY3VYOP:-MYAa9K8AZ2uOyY3VYOQ:b-r4h2hr) ([raw file](https://github.com/m-lab/etl/blob/0e06744425e7234f9d1a37f84d42275767bd06f4/site/site.go#L59)):*
> ```Go
> // Load loads the site annotations source. Will try at least once, retry up to
> // timeout and return an error if unsuccessful.
> func Load(timeout time.Duration) error {
> ```

It may be helpful to accept a ctx here for the reload loop goroutine, and have the caller able to cancel the ctx. I don't think that's required here, but might be a more consistent use of ctx's.
___
*[site/site.go, line 74 at r3](https://reviewable.io/reviews/m-lab/etl/982#-MYA__Kb38SklxM0KSj5:-MYA__Kb38SklxM0KSj6:b-azlbnm) ([raw file](https://github.com/m-lab/etl/blob/0e06744425e7234f9d1a37f84d42275767bd06f4/site/site.go#L74)):*
> ```Go
> 
> 	for ; ctx.Err() == nil; time.Sleep(time.Second) {
> 		err = LoadFrom(context.Background(), js, retiredJS)
> ```

Should this use the `ctx` with the timeout?

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/982)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @robertodauria and @stephen-soltesz)

___
*[site/site.go, line 74 at r3](https://reviewable.io/reviews/m-lab/etl/982#-MYA__Kb38SklxM0KSj5:-MYAd_dIEU90hxxIRyBN:bjlwgg1) ([raw file](https://github.com/m-lab/etl/blob/0e06744425e7234f9d1a37f84d42275767bd06f4/site/site.go#L74)):*

Should this use the `ctx` with the timeout?
</blockquote></details>

It would break gcpProvider by giving `Get(ctx)` a longer-than-intended timeout, making the retry loop useless as it would always try just once. `httpProvider` would not be much affected as it creates its own `context.WithTimeout(ctx, time.Minute)` from the context that's passed in. If we are sure this will always be an `httpProvider` (i.e. we are never moving to a `gs://` scheme for the two URLs), it could also be fine to pass `ctx` rather than `context.Background()` there, but it felt a bit wrong to me to break one of the possible implementations. Right now that ctx is only used for the retry loop. Probably the behavior of `gcpProvider.Get(ctx)` and `httpProvider.Get(ctx)` should be standardized a bit.

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/982)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[site/site.go, line 54 at r3](https://reviewable.io/reviews/m-lab/etl/982#-MYAZ_Sr3sbJfHvfejGD:-MYAi5oB3jgxpTYd-yKY:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/0e06744425e7234f9d1a37f84d42275767bd06f4/site/site.go#L54)):*

Please clarify "server annotations". Including the intended URLs might be helpful too. Depending on the error, the err string might not be enough.
</blockquote></details>

Done.

stephen-soltesz (approved):
:lgtm: -- just a thought, if this code does need to be linked into the annotation-service client library, this code might live more naturally in the annotation-service repo.

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/982#-:-MYB0_wkBQrIR_SFqqRf:bbr6mq3)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @robertodauria)

___
*[site/site.go, line 74 at r3](https://reviewable.io/reviews/m-lab/etl/982#-MYA__Kb38SklxM0KSj5:-MYB-9oY-ix6iyKan6mL:b78y1yu) ([raw file](https://github.com/m-lab/etl/blob/0e06744425e7234f9d1a37f84d42275767bd06f4/site/site.go#L74)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

It would break gcpProvider by giving `Get(ctx)` a longer-than-intended timeout, making the retry loop useless as it would always try just once. `httpProvider` would not be much affected as it creates its own `context.WithTimeout(ctx, time.Minute)` from the context that's passed in. If we are sure this will always be an `httpProvider` (i.e. we are never moving to a `gs://` scheme for the two URLs), it could also be fine to pass `ctx` rather than `context.Background()` there, but it felt a bit wrong to me to break one of the possible implementations. Right now that ctx is only used for the retry loop. Probably the behavior of `gcpProvider.Get(ctx)` and `httpProvider.Get(ctx)` should be standardized a bit.
</blockquote></details>

Would you create an issue for standardizing the provider behaviors for `Get(ctx)`? Requiring background contexts for `FromURL` and the loop behavior is a weird pattern to require callers to use.

There are two failure modes that we want to be able to handle: hanging connections, transient failures. Hanging is a failure case that should be time-limited, but I think of the common case being short transient failures that a quick retry would resolve.

robertodauria:
https://github.com/m-lab/go/issues/130

PR #998, opened 2021-06-28
Add synthetic UUIDs for legacy datatypes

stephen-soltesz:
This change adds new `id` columns to the legacy datatypes for ndt.web100 and sidestream and reuses the `UUID` column for traceroute1 (.paris).

The id is a synthetic UUID, derived from the same data used by gardener to dedup daily columns. Because all of these include dates, they are considered globally unique. The id also relies on the MD5 digest. The only collisions observed from manual inspection of these BQ tables come from duplicate data in archives uploaded to multiple archives (so the "collisions" are correctly identifying the same data).

These ids can be used directly during the annotation export process to generate synthetic uuid-annotations.

These ids will be preserved indefinitely to maintain compatibility with the synthetic uuid-annotations.

Part of milestone: https://docs.google.com/document/d/1seI56IGAZzfIhmkZH_Pp67fU11kynyO6mwf7gU3HeiM/edit#

This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/998)


gfr10598 (approved):
:lgtm:  after considering some suggestions.

> Reviewed 8 of 9 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998#-:-MdJgS498p17HPsVy0V7:bn7qv0f)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @robertodauria and @stephen-soltesz)

___
*[parser/parser.go, line 135 at r2](https://reviewable.io/reviews/m-lab/etl/998#-MdJeiyi4fATEDdndpWQ:-MdJeiyi4fATEDdndpWR:b-4dutq9) ([raw file](https://github.com/m-lab/etl/blob/e545b5472e0f15bb7330ce6d5e3031bd51265f08/parser/parser.go#L135)):*
> ```Go
> 
> // syntheticUUID produces an opague, filename-safe string. The resulting string
> // should only be used to provide uniqueness. Strings are fixed length.
> ```

Suggest updating the comment to:
// syntheticUUID produces an opaque, filename-safe string. The resulting string
// should only be used to provide unique identifiers for joining with annotations.

notice corrected typo for opague.

Alternatively, perhaps state this at the points where this is called, since it is the string passed in that determines the utility.
___
*[parser/parser.go, line 136 at r2](https://reviewable.io/reviews/m-lab/etl/998#-MdJfwH7-psHM_V1E0SH:-MdJfwH7-psHM_V1E0SI:b-1x2gjh) ([raw file](https://github.com/m-lab/etl/blob/e545b5472e0f15bb7330ce6d5e3031bd51265f08/parser/parser.go#L136)):*
> ```Go
> // syntheticUUID produces an opague, filename-safe string. The resulting string
> // should only be used to provide uniqueness. Strings are fixed length.
> func syntheticUUID(input string) string {
> ```

Hmm - rename this something like base64hash?  It has nothing in it that is specific to UUIDs.
___
*[parser/parser.go, line 137 at r2](https://reviewable.io/reviews/m-lab/etl/998#-MdJe-V02p4dkKjL-SAw:-MdJe-V02p4dkKjL-SAx:bbdqilm) ([raw file](https://github.com/m-lab/etl/blob/e545b5472e0f15bb7330ce6d5e3031bd51265f08/parser/parser.go#L137)):*
> ```Go
> // should only be used to provide uniqueness. Strings are fixed length.
> func syntheticUUID(input string) string {
> 	b := sha1.Sum([]byte(input))
> ```

Didn't register when we discussed earlier, but it might be better to use MD5, just because I think it is more often used for this kind of thing - faster, and less mistaken for being cryptographically secure.
Not a show stopper though.

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @gfr10598 and @robertodauria)

___
*[parser/parser.go, line 135 at r2](https://reviewable.io/reviews/m-lab/etl/998#-MdJeiyi4fATEDdndpWQ:-MdJqKm03OiU9kEBAOau:bq5sa7r) ([raw file](https://github.com/m-lab/etl/blob/e545b5472e0f15bb7330ce6d5e3031bd51265f08/parser/parser.go#L135)):*

Suggest updating the comment to:
// syntheticUUID produces an opaque, filename-safe string. The resulting string
// should only be used to provide unique identifiers for joining with annotations.

notice corrected typo for opague.

Alternatively, perhaps state this at the points where this is called, since it is the string passed in that determines the utility.

</blockquote></details>

That the uniqueness is a function of the input is a good point. I've renamed this function to be generic and defined helper functions for the specific datatypes. Each function is now narrowly scoped, clearly defined, and not overloaded.
___
*[parser/parser.go, line 136 at r2](https://reviewable.io/reviews/m-lab/etl/998#-MdJfwH7-psHM_V1E0SH:-MdJoVPE-EKEeWvTOK67:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/e545b5472e0f15bb7330ce6d5e3031bd51265f08/parser/parser.go#L136)):*

Hmm - rename this something like base64hash?  It has nothing in it that is specific to UUIDs.
</blockquote></details>

Done.
___
*[parser/parser.go, line 137 at r2](https://reviewable.io/reviews/m-lab/etl/998#-MdJe-V02p4dkKjL-SAw:-MdJqlW782ArhGift4Aa:bd4dtma) ([raw file](https://github.com/m-lab/etl/blob/e545b5472e0f15bb7330ce6d5e3031bd51265f08/parser/parser.go#L137)):*

Didn't register when we discussed earlier, but it might be better to use MD5, just because I think it is more often used for this kind of thing - faster, and less mistaken for being cryptographically secure.
Not a show stopper though.
</blockquote></details>

Good question. Since we'll be doing this trillions of times, this may be better consideration. I'll sleep on it b/c while I think md5 had the same no-collisions property as sha1, I didn't try as hard and would need to double check my work again with md5. The tests and examples would be easy enough to update.

stephen-soltesz (commented):
@robertodauria PTAL also. This change is relevant for the export process.

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998#-:-MdJrsOu7NMsj7uwzylo:b1qnyqk)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @gfr10598 and @robertodauria)

robertodauria (approved):
:lgtm:

> Reviewed 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998#-:-MdMsEmr99ApFR2je8HF:bnfp4nl)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @gfr10598)

stephen-soltesz (commented):
I've updated the hash to use md5. I double checked that there are no collisions. All "duplicate" hashes are from the same file in two independent archives.

See: https://console.cloud.google.com/bigquery?sq=581276032543:c47c31e5ad0945b79de9dfac6dc9e52e 

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998#-:-MdOmYyX8EUOpsXsvkta:bovysbb)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @gfr10598)

stephen-soltesz (commented):
Also, I found that the sidestream synthetic UUID function was missing the file name, which gardener uses in the dedup.

PTAL?

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998#-:-MdOnb_Z2Zq0ccbaMjkm:b-x39siq)* status: :shipit: complete! 2 of 1 approvals obtained (waiting on @gfr10598)

___
*[parser/parser.go, line 137 at r2](https://reviewable.io/reviews/m-lab/etl/998#-MdJe-V02p4dkKjL-SAw:-MdOnZUc6lw0xt_UGSlS:b1qs466) ([raw file](https://github.com/m-lab/etl/blob/e545b5472e0f15bb7330ce6d5e3031bd51265f08/parser/parser.go#L137)):*

Good question. Since we'll be doing this trillions of times, this may be better consideration. I'll sleep on it b/c while I think md5 had the same no-collisions property as sha1, I didn't try as hard and would need to double check my work again with md5. The tests and examples would be easy enough to update.
</blockquote></details>

I've updated the hash to use md5.

gfr10598 (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998)* status: :shipit: complete! 2 of 1 approvals obtained

___
*[parser/parser.go, line 137 at r2](https://reviewable.io/reviews/m-lab/etl/998#-MdJe-V02p4dkKjL-SAw:-MdOwXlB1N9S9yCQ_9I0:bfehkkc) ([raw file](https://github.com/m-lab/etl/blob/e545b5472e0f15bb7330ce6d5e3031bd51265f08/parser/parser.go#L137)):*

I've updated the hash to use md5.
</blockquote></details>

FYI, MD5 and sha1 both have the possibility of collisions, and we can only ensure that they will be rare.  (I guess some poor hashes are more likely to create collisions on short strings, but I think md5 and sha1 are both pretty good?). AFAICT we really only care about uniqueness within a date partition, and even if we only used 64 bit hashes, collisions within partitions would be very unlikely.  With 128 bit hash, it will be extremely unlikely to see a collision across the entire archive history.

gfr10598 (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998)* status: :shipit: complete! 2 of 1 approvals obtained

___
*[parser/pt.go, line 619 at r4](https://reviewable.io/reviews/m-lab/etl/998#-MdOxu3b4lGElLXjydvy:-MdOxu3cB-kwFH-OjGaz:b-qctwqk) ([raw file](https://github.com/m-lab/etl/blob/2b6afcfa806635abbf22af6cb67cbb9784c94b98/parser/pt.go#L619)):*
> ```Go
> 	}
> 
> 	// Since this is a .paris file, we create a synthetic UUID for joining with annotations.
> ```

I don't really understand the intent of this comment.

gfr10598 (approved):
added a couple more comments, then :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998#-:-MdOy4I6CmorgcCPeL-G:bf3dhs8)* status: :shipit: complete! 2 of 1 approvals obtained

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998)* status: :shipit: complete! 2 of 1 approvals obtained

___
*[parser/parser.go, line 137 at r2](https://reviewable.io/reviews/m-lab/etl/998#-MdJe-V02p4dkKjL-SAw:-MdSBRyXDPHV8_MrGui8:b-teuicz) ([raw file](https://github.com/m-lab/etl/blob/e545b5472e0f15bb7330ce6d5e3031bd51265f08/parser/parser.go#L137)):*

FYI, MD5 and sha1 both have the possibility of collisions, and we can only ensure that they will be rare.  (I guess some poor hashes are more likely to create collisions on short strings, but I think md5 and sha1 are both pretty good?). AFAICT we really only care about uniqueness within a date partition, and even if we only used 64 bit hashes, collisions within partitions would be very unlikely.  With 128 bit hash, it will be extremely unlikely to see a collision across the entire archive history.
</blockquote></details>

Yes, you mentioned the potential for collision a month or so ago. I felt it was extremely low probability but still wanted to establish that it would provide a globally unique ID (b/c that's what people expect from UUID) for our static, historical datatypes and ran the BQ queries to establish that the only "duplicate" hashes come from duplicate files in independent archives. This should be consistent with the [design](https://docs.google.com/document/d/1Hdf7-FglNs5OXMBcAbO_b0HZZxP-4nkXPXnZwDlUEII/edit?resourcekey=0-eun68veHrU1Y0aNUJpAZdg#) - please let me know if there's anything out of sync.

Though, I see I did not replicate the input strings precisely in the BQ checks, so I'll do that again before merging.
See: https://console.cloud.google.com/bigquery?sq=581276032543:c47c31e5ad0945b79de9dfac6dc9e52e
___
*[parser/pt.go, line 619 at r4](https://reviewable.io/reviews/m-lab/etl/998#-MdOxu3b4lGElLXjydvy:-MdSSCrnBqONcVdfRWeE:b-ms1n3e) ([raw file](https://github.com/m-lab/etl/blob/2b6afcfa806635abbf22af6cb67cbb9784c94b98/parser/pt.go#L619)):*

I don't really understand the intent of this comment.
</blockquote></details>

I don't understand your comment. Can you say more?

Each invocation of the synthetic UUID for other datatypes has a similar comment. For .paris files, only the traceroute1 datatype needs synthetic UUIDs, traceroute2++ use real UUIDs.

stephen-soltesz (commented):
I think this PR is complete. I've rerun the bq validation using equivalent input strings for the md5 ids and still found no collisions.

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/998#-:-MdSYNgAEwg46_J3_Xuf:ba9gtgm)* status: :shipit: complete! 2 of 1 approvals obtained

PR #1012, opened 2021-07-25
Add parser for pcap index

cristinaleonr:
- Adding parser for pcap datatype
- Adding pcap datatype constants

Sample output
```
{
  "ID":"ndt-4c6fb_1625899199_00000000013A461F",
  "Parser":  {
    "Version":"local development",
    "Time":"2021-07-25T17:01:08.974024413Z",
    "ArchiveURL":"gs://archive-measurement-lab/ndt/pcap/2021/07/22/20210722T000107.470279Z-pcap-mlab1-dfw05-ndt.tgz",
    "Filename":"2021/07/22/ndt-4c6fb_1625899199_00000000013A461F.pcap.gz",
    "Priority":0,
    "GitCommit":"uninitialized"
  },
  "Date":"2021-07-22"
}
```


stephen-soltesz (changes requested):
> Reviewed 5 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1012)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[etl/globals.go, line 306 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZEfc50byLv0mx9mlC:-MfZEfc50byLv0mx9mlD:bklqd5r) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/etl/globals.go#L306)):*
> ```Go
> 	// as everything is written to gcs files, and the gcs library does it’s own buffering.
> 	// TODO - this should be loaded from a config
> 	dataTypeToBQBufferSize = map[DataType]int{
> ```

@gfr10598 do we need these defined for standard-column, gardener 2.0 types at all?
___
*[parser/pcap.go, line 44 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZFjlE8GmN25dTR7uW:-MfZFjlE8GmN25dTR7uX:bppfn9i) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap.go#L44)):*
> ```Go
> 
> // IsParsable returns the canonical test type and whether to parse data.
> func (parser *PCAPParser) IsParsable(testName string, data []byte) (string, bool) {
> ```

In this case I would recommend simply using `p` (or other short name) for the receiver name instead of `parser`. In Go there is a "tradition" (not sure the right word; it's not style) of using short names for receivers. In this case, `parser` is also the name of the package and could be accidentally confusing.
___
*[parser/pcap.go, line 89 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZI46o-jXLuoXls6gx:-MfZI46o-jXLuoXls6gy:b-b14ops) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap.go#L89)):*
> ```Go
> // it returns ndt-4c6fb_1625899199_00000000013A4623.
> func (parser *PCAPParser) GetUUID(filename string) string {
> 	regex := regexp.MustCompile(`\d{4}/\d{2}/\d{2}/`)
> ```

There are times when I'm happy, and there are times when I need to use regexp.

If we can avoid regexp for simple cases, I believe this will make the logic clearer in the long run. In this case, we could use a combination of `filepath.Base` to extract the base file name, and then `strings.TrimSuffix` using the same constant string used to identify the file in `IsParseable`.

This would eliminate the need for regexp and remove a magic number `8`.
___
*[parser/pcap_test.go, line 17 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZKdamEL5gpRz1bSHR:-MfZKdamEL5gpRz1bSHS:bykrlzh) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap_test.go#L17)):*
> ```Go
> 
> const (
> 	filename = "ndt-4c6fb_1625899199_000000000121C1A0.pcap.gz"
> ```

Defining these constants is good.

We have no convention for this, so there are probably counter examples already in our codebase -- since Go package tests also share a namespace, it may be helpful to use less generic constant names. e.g. pcapFilename, pcapGCSPath.
___
*[parser/pcap_test.go, line 25 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZJp2rCXEF3-kJF2tG:-MfZJp2rCXEF3-kJF2tH:b-7rpvux) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap_test.go#L25)):*
> ```Go
> 	n := parser.NewPCAPParser(ins, "test", "_suffix", &fakeAnnotator{})
> 
> 	data, err := ioutil.ReadFile("testdata/PCAP/" + filename)
> ```

I'm sure this is not done consistently in our codebase. In cases like this, rather than our relying on well formed prefixes with trailing "/" suffixes, I would encourage the use of `path.Join`, which ensures that the path components are combined correctly. So, in this case, something like:

```
ioutil.ReadFile(path.Join("testdata/PCAP", filename))
```
___
*[parser/pcap_test.go, line 49 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZKF-OBtMDEXfI5XWF:-MfZKF-OBtMDEXfI5XWG:b3dw69k) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap_test.go#L49)):*
> ```Go
> 		Version:    "https://github.com/m-lab/etl/tree/foobar",
> 		Time:       row.Parser.Time,
> 		ArchiveURL: GCSPath + filename,
> ```

I think it's possible to use `path.Join` here also (and anywhere else two path components are combined.)

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1012)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr and @stephen-soltesz)

___
*[parser/pcap.go, line 44 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZFjlE8GmN25dTR7uW:-MfZmch4DwffqAI6HTW0:bpcxg5x) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap.go#L44)):*

In this case I would recommend simply using `p` (or other short name) for the receiver name instead of `parser`. In Go there is a "tradition" (not sure the right word; it's not style) of using short names for receivers. In this case, `parser` is also the name of the package and could be accidentally confusing.
</blockquote></details>

Changed.

I was surprised to see that one/two-letter names are not just the convention, but the actual style guideline https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#receiver-names
___
*[parser/pcap.go, line 89 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZI46o-jXLuoXls6gx:-MfZnLZcEgZTne2bmkJn:bcwb8xt) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap.go#L89)):*

There are times when I'm happy, and there are times when I need to use regexp.

If we can avoid regexp for simple cases, I believe this will make the logic clearer in the long run. In this case, we could use a combination of `filepath.Base` to extract the base file name, and then `strings.TrimSuffix` using the same constant string used to identify the file in `IsParseable`.

This would eliminate the need for regexp and remove a magic number `8`.
</blockquote></details>

Thanks, this is much better.
___
*[parser/pcap_test.go, line 17 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZKdamEL5gpRz1bSHR:-MfZnSpr07xA7zYHF3RI:bctpdcc) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap_test.go#L17)):*

Defining these constants is good.

We have no convention for this, so there are probably counter examples already in our codebase -- since Go package tests also share a namespace, it may be helpful to use less generic constant names. e.g. pcapFilename, pcapGCSPath.
</blockquote></details>

Makes sense. Done.
___
*[parser/pcap_test.go, line 25 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZJp2rCXEF3-kJF2tG:-MfZnVl37VRA6EO4ULkx:b9ig8r8) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap_test.go#L25)):*

I'm sure this is not done consistently in our codebase. In cases like this, rather than our relying on well formed prefixes with trailing "/" suffixes, I would encourage the use of `path.Join`, which ensures that the path components are combined correctly. So, in this case, something like:

```
ioutil.ReadFile(path.Join("testdata/PCAP", filename))
```
</blockquote></details>

Great. Changed.
___
*[parser/pcap_test.go, line 49 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZKF-OBtMDEXfI5XWF:-MfZn_AcD16KwDH7QHni:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap_test.go#L49)):*

I think it's possible to use `path.Join` here also (and anywhere else two path components are combined.)
</blockquote></details>

Done.

gfr10598 (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1012)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[etl/globals.go, line 306 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZEfc50byLv0mx9mlC:-MfcIJwc7JdZotzV0QA2:bqlrgid) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/etl/globals.go#L306)):*

@gfr10598 do we need these defined for standard-column, gardener 2.0 types at all?
</blockquote></details>

No, but they are currently part of the plumbing in row.NewBase.  Hopefully can remove them later, but best to leave it as is for now.

stephen-soltesz (approved):
:lgtm:

> Reviewed 2 of 8 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1012#-:-MfcKaPsEjH_gL3PhLP_:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @SaiedKazemi)

___
*[parser/pcap.go, line 44 at r1](https://reviewable.io/reviews/m-lab/etl/1012#-MfZFjlE8GmN25dTR7uW:-MfcJbP94wS64w7eFfYE:byaremc) ([raw file](https://github.com/m-lab/etl/blob/108c88d6bc14193066bea6105d8416a126c81e64/parser/pcap.go#L44)):*

Changed.

I was surprised to see that one/two-letter names are not just the convention, but the actual style guideline https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#receiver-names
</blockquote></details>

Ah, yes, good. I'm glad you found that. @saiedkazemi is working on making resources like that link publicly accessible.

PR #1014, opened 2021-08-12
Update pcap parser processing rate to 1 in 10 archives

cristinaleonr:
Update pcap parser to process fewer archives (i.e., 1 in 10)

A similar concept exists in the legacy pipeline with the TASK_FILE_SKIP setting (e.g., https://github.com/m-lab/etl-gardener/blob/master/k8s/data-processing-cluster/deployments/etl-gardener-ndt.yml#L52)

Testing comparing the processing the logs to the archive entries

Log
https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22mlab-sandbox%22%0Aresource.labels.location%3D%22us-east1%22%0Aresource.labels.cluster_name%3D%22data-processing%22%0Aresource.labels.namespace_name%3D%22default%22%0Alabels.k8s-pod%2Frun%3D%22etl-parser%22%20severity%3E%3DDEFAULT;timeRange=2021-08-12T17:13:17.000Z%2F2021-08-12T17:13:18.000Z;cursorTimestamp=2021-08-12T17:13:17.602104577Z?project=mlab-sandbox
Screenshot 2021-08-12 1 46 21 PM

Archive
https://pantheon.corp.google.com/storage/browser/archive-measurement-lab/ndt/pcap/2021/07/29?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&project=measurement-lab&prefix=&forceOnObjectsSortingFiltering=false
Screenshot 2021-08-12 2 07 20 PM

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/1014)

### DIFF
--- a/active/active.go
+++ b/active/active.go
@@ -20,6 +20,8 @@ import (
 	"google.golang.org/api/iterator"
 
 	"cloud.google.com/go/storage"
+	"github.com/m-lab/etl-gardener/tracker"
+	"github.com/m-lab/etl/etl"
 	"github.com/m-lab/etl/metrics"
 	"github.com/m-lab/go/cloud/gcs"
 	"github.com/m-lab/go/logx"
@@ -103,17 +105,17 @@ type GCSSource struct {
 }
 
 // NewGCSSource creates a new source for active processing.
-func NewGCSSource(ctx context.Context, label string, fl FileLister, toRunnable func(*storage.ObjectAttrs) Runnable) (*GCSSource, error) {
+func NewGCSSource(ctx context.Context, job tracker.Job, fl FileLister, toRunnable func(*storage.ObjectAttrs) Runnable) (*GCSSource, error) {
 	src := GCSSource{
 		ctx:        WithFail(ctx),
 		fileLister: fl,
 		toRunnable: toRunnable,
 
 		pendingChan: make(chan Runnable, 0),
-		label:       label,
+		label:       job.Path(),
 	}
 
-	go src.streamToPending(ctx)
+	go src.streamToPending(ctx, job)
 
 	return &src, nil
 }
@@ -162,7 +164,7 @@ func (src *GCSSource) Next(ctx context.Context) (Runnable, error) {
 // It fetches the list of files once, then converts files to Runnables until all files are
 // handled, or the context is canceled or expires.
 // The Runnables are pulled from the queue by Next().
-func (src *GCSSource) streamToPending(ctx context.Context) {
+func (src *GCSSource) streamToPending(ctx context.Context, job tracker.Job) {
 	// No matter what else happens, we eventually want to close the pendingChan.
 	defer close(src.pendingChan)
 
@@ -173,6 +175,10 @@ func (src *GCSSource) streamToPending(ctx context.Context) {
 		src.ctx.Fail(err)
 		return
 	}
+
+	index := 0
+	dataType := etl.NameToDataType(job.Datatype)
+	skipCount := dataType.SkipCount()
 
 	for _, f := range files {
 		debug.Println(f)
@@ -186,8 +192,12 @@ func (src *GCSSource) streamToPending(ctx context.Context) {
 			metrics.ActiveErrors.WithLabelValues(src.Label(), "streamToPending").Inc()
 			break
 		}
-		debug.Printf("Adding gs://%s/%s", f.Bucket, f.Name)
-		// Blocks until consumer reads channel.
-		src.pendingChan <- src.toRunnable(f)
+
+		if index%(skipCount+1) == 0 {
+			debug.Printf("Adding gs://%s/%s", f.Bucket, f.Name)
+			// Blocks until consumer reads channel.
+			src.pendingChan <- src.toRunnable(f)
+		}
+		index++
 	}
 }
--- a/active/active.go
+++ b/active/active.go
@@ -177,7 +177,7 @@ func (src *GCSSource) streamToPending(ctx context.Context, job tracker.Job) {
 	}
 
 	index := 0
-	dataType := etl.NameToDataType(job.Datatype)
+	dataType := etl.DataType(job.Datatype)
 	skipCount := dataType.SkipCount()
 
 	for _, f := range files {
--- a/active/active_test.go
+++ b/active/active_test.go
@@ -109,7 +109,7 @@ func standardLister() active.FileLister {
 
 func skipFilesListener(dataType string) active.FileLister {
 	client := gcsfake.GCSClient{}
-	bucket := "ndt/" + dataType + "/2019/01/01/"
+	bucket := path.Join("ndt/", dataType, "/2019/01/01/")
 	client.AddTestBucket("foobar",
 		&gcsfake.BucketHandle{
 			ObjAttrs: []*storage.ObjectAttrs{
--- a/active/active_test.go
+++ b/active/active_test.go
@@ -109,26 +109,26 @@ func standardLister() active.FileLister {
 
 func skipFilesListener(dataType string) active.FileLister {
 	client := gcsfake.GCSClient{}
-	bucket := path.Join("ndt/", dataType, "/2019/01/01/")
+	prefix := path.Join("ndt/", dataType, "/2019/01/01/")
 	client.AddTestBucket("foobar",
 		&gcsfake.BucketHandle{
 			ObjAttrs: []*storage.ObjectAttrs{
-				{Bucket: "foobar", Name: path.Join(bucket, "obj1"), Updated: time.Now()},
-				{Bucket: "foobar", Name: path.Join(bucket, "obj2"), Updated: time.Now()},
-				{Bucket: "foobar", Name: path.Join(bucket, "obj3"), Updated: time.Now()},
-				{Bucket: "foobar", Name: path.Join(bucket, "obj4"), Updated: time.Now()},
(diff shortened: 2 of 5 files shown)


stephen-soltesz (changes requested):
Excellent! Thank you very much for testing in sandbox to verify that the system works as intended. Thank you for the helpful screenshots to confirm.

I've added a few comments that could help simplify. Let me know what you think.

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1014#-:-MgveF6R63FWVwDAkrLG:b-sqheiv)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[active/active.go, line 180 at r1](https://reviewable.io/reviews/m-lab/etl/1014#-Mgve6H1CqpGmyEJ_xAO:-Mgvbh94-GCTeSARp6dP:b-c9va7w) ([raw file](https://github.com/m-lab/etl/blob/a9948c2174425263d75b95f2ab53fdaad713f8d3/active/active.go#L180)):*
> ```Go
> 
> 	index := 0
> 	dataType := etl.NameToDataType(job.Datatype)
> ```

This could be a cast: `etl.DataType(job.Datatype)`
___
*[etl/globals.go, line 325 at r1](https://reviewable.io/reviews/m-lab/etl/1014#-MgvaKO78B8_hWAi24jh:-MgvaKO78B8_hWAi24ji:b-24um8v) ([raw file](https://github.com/m-lab/etl/blob/a9948c2174425263d75b95f2ab53fdaad713f8d3/etl/globals.go#L325)):*
> ```Go
> 	// It allows us process fewer archives when there is a very high volume of data.
> 	// TODO - this should be loaded from a config.
> 	dataTypeToSkipCount = map[DataType]int{
> ```

Ah, nice. A behavior of Go maps is that when a key is not present, a lookup returns the default zero-value for the the value type. In this case, the zero-value for an int is zero. If I used a construct like:

```
fmt.Println(dataTypeToSkipCount[DataType("fakedatatype")])
```

It would print zero.

Because you're using `index%(skipCount+1) == 0` to check for skips, I think it would be sufficient here to leave all datatypes unspecified except for those that need a skip count > 0. What do you think?

The implementation of `SkipCount` becomes one line.
___
*[etl/globals.go, line 351 at r1](https://reviewable.io/reviews/m-lab/etl/1014#-MgvdEIv8wYw8FOKMlVB:-MgvdEIv8wYw8FOKMlVC:b-u6pabm) ([raw file](https://github.com/m-lab/etl/blob/a9948c2174425263d75b95f2ab53fdaad713f8d3/etl/globals.go#L351)):*
> ```Go
> 
> // NameToDataType returns the DataType corresponding to the type name.
> func NameToDataType(name string) DataType {
> ```

Do we need this function? Could callers use `etl.DataType` to cast names on demand?

```
dt := etl.DataType(job.Datatype)
skip := dt.SkipCount()
```

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1014)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[active/active.go, line 180 at r1](https://reviewable.io/reviews/m-lab/etl/1014#-Mgve6H1CqpGmyEJ_xAO:-Mgvi3ovADFmOiFZWsJK:b3px6hb) ([raw file](https://github.com/m-lab/etl/blob/a9948c2174425263d75b95f2ab53fdaad713f8d3/active/active.go#L180)):*

This could be a cast: `etl.DataType(job.Datatype)`
</blockquote></details>

Thanks, changed!
___
*[etl/globals.go, line 325 at r1](https://reviewable.io/reviews/m-lab/etl/1014#-MgvaKO78B8_hWAi24jh:-Mgvi5Xy7V_9gd-eCISW:b-4pj5my) ([raw file](https://github.com/m-lab/etl/blob/a9948c2174425263d75b95f2ab53fdaad713f8d3/etl/globals.go#L325)):*

Ah, nice. A behavior of Go maps is that when a key is not present, a lookup returns the default zero-value for the the value type. In this case, the zero-value for an int is zero. If I used a construct like:

```
fmt.Println(dataTypeToSkipCount[DataType("fakedatatype")])
```

It would print zero.

Because you're using `index%(skipCount+1) == 0` to check for skips, I think it would be sufficient here to leave all datatypes unspecified except for those that need a skip count > 0. What do you think?

The implementation of `SkipCount` becomes one line.
</blockquote></details>

Nice, done.
___
*[etl/globals.go, line 351 at r1](https://reviewable.io/reviews/m-lab/etl/1014#-MgvdEIv8wYw8FOKMlVB:-Mgvi9Hf6SKxths-9zCj:bss4a2b) ([raw file](https://github.com/m-lab/etl/blob/a9948c2174425263d75b95f2ab53fdaad713f8d3/etl/globals.go#L351)):*

Do we need this function? Could callers use `etl.DataType` to cast names on demand?

```
dt := etl.DataType(job.Datatype)
skip := dt.SkipCount()
```
</blockquote></details>

Removed.

stephen-soltesz (changes requested):
> Reviewed 1 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1014)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[active/active_test.go, line 293 at r2](https://reviewable.io/reviews/m-lab/etl/1014#-Mgw1UP81_t83gkRxraD:-Mgw1UP81_t83gkRxraE:b-wwbzzs) ([raw file](https://github.com/m-lab/etl/blob/cae968028e92a0bfdfea65986b2d96538949f9f0/active/active_test.go#L293)):*
> ```Go
> 			p := newCounter(t)
> 			ctx := context.Background()
> 			fs, err := active.NewGCSSource(ctx, tracker.Job{Datatype: tt.name}, standardLister(testClientSkipFiles), p.toRunnable)
> ```

I have mixed feelings about the `testClientSkipFiles` statically encoding the "ndt5" datatype in the obj paths, but the test names here ("pcap", "ndt7", "foo") are treated as the datatype.

If the obj paths are not meaningful for the code under test, then I would recommend using something like "fake" instead of a real datatype name.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1014)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[active/active_test.go, line 293 at r2](https://reviewable.io/reviews/m-lab/etl/1014#-Mgw1UP81_t83gkRxraD:-MgzzUskCE7UDok76DKU:bpllhfs) ([raw file](https://github.com/m-lab/etl/blob/cae968028e92a0bfdfea65986b2d96538949f9f0/active/active_test.go#L293)):*

I have mixed feelings about the `testClientSkipFiles` statically encoding the "ndt5" datatype in the obj paths, but the test names here ("pcap", "ndt7", "foo") are treated as the datatype.

If the obj paths are not meaningful for the code under test, then I would recommend using something like "fake" instead of a real datatype name.
</blockquote></details>

We do have some validation for the path, so adding "fake" does not work in this case. 

I've added the actual datatypes tested.

stephen-soltesz (approved):
One variable name change, and :lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1014#-:-Mh-8bw9EydEXpJI98ku:b-tkrp9z)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[active/active_test.go, line 293 at r2](https://reviewable.io/reviews/m-lab/etl/1014#-Mgw1UP81_t83gkRxraD:-Mh-8LhF30ybOgf_iJOP:b-qn5lay) ([raw file](https://github.com/m-lab/etl/blob/cae968028e92a0bfdfea65986b2d96538949f9f0/active/active_test.go#L293)):*

We do have some validation for the path, so adding "fake" does not work in this case. 

I've added the actual datatypes tested.
</blockquote></details>

Perfect.
___
*[active/active_test.go, line 112 at r3](https://reviewable.io/reviews/m-lab/etl/1014#-Mh-86kk2eQIXifitgVZ:-Mh-86kk2eQIXifitgV_:b-zcp3wb) ([raw file](https://github.com/m-lab/etl/blob/a82d08367a58671274b274e4cdaedbdf561d76fb/active/active_test.go#L112)):*
> ```Go
> func skipFilesListener(dataType string) active.FileLister {
> 	client := gcsfake.GCSClient{}
> 	bucket := path.Join("ndt/", dataType, "/2019/01/01/")
> ```

Please call this "prefix". In the GCS context the "bucket" is the term to describe the thing that contains all of the named objects inside the bucket. The path prefix here is part of naming the full paths for obj1-11.

cristinaleonr:
___
*[active/active_test.go, line 112 at r3](https://reviewable.io/reviews/m-lab/etl/1014#-Mh-86kk2eQIXifitgVZ:-Mh-9iZw6lxfFE_IhAi1:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/a82d08367a58671274b274e4cdaedbdf561d76fb/active/active_test.go#L112)):*

Please call this "prefix". In the GCS context the "bucket" is the term to describe the thing that contains all of the named objects inside the bucket. The path prefix here is part of naming the full paths for obj1-11.
</blockquote></details>

Done.

PR #1034, opened 2021-11-30
Tweak slice allocation to improve parsing performance.

gfr10598:
This is a minor tweak, but useful to get a better baseline before adding a lot more code.

It also adds a better benchmark, that includes accounting for the packets/op, so that it is easier to estimate time and memory allocated per packet.


stephen-soltesz:
Closing old PR - and referencing the notes from this change in https://github.com/m-lab/etl/issues/1085

PR #1036, opened 2021-12-13
fix file size bins

gfr10598:
When using heatmaps in Grafana, non-uniform bins cause visual artifacts that can be quite misleading, or hide actual features of significance.  The important feature is that the bin distribution should be smooth, i.e. either linear, logarithmic, polynomial, etc.

We have a few hand crafted bin sets in our metrics that are not very uniform (neither linear nor logarithmic).  This PR fixes one of those that was causing gfr confusion in mlab-sandbox development.

metrics.FileSizeHistogram will now have a smooth bin distribution, and easier interpretation.


gfr10598:
Before
Screen Shot 2021-12-13 at 7 44 38 AM
After
Screen Shot 2021-12-13 at 7 43 42 AM

stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1036)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1042, opened 2021-12-23
Add script to delete extra appengine services

stephen-soltesz:
AppEngine has a hard limit of 210 service versions across a project. This change adds a small script that deletes all service versions after the first 9 that are non-serving. The number of older versions to preserve is arbitrary; more than 1 to allow rollbacks is the main intention. 9 is conservative number. This change was inspired by an alert in staging about there being too many service versions.

Limitations / caveats:

- The first time this script runs, it will delete many older versions and will take longer than steady state. At the limit, it should delete no more than one per service.
- This script applies to all app engine services but will only run on ETL deployments.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/1042)

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -140,3 +140,9 @@ steps:
   args: [
     'bash', '-c', 'cd cmd/etl_worker && gcloud --project=$PROJECT_ID app deploy --promote app-batch.yaml'
   ]
+
+- name: gcr.io/google.com/cloudsdktool/cloud-sdk
+  id: "Delete non-serving appengine versions"
+  args: [
+    'bash', '-c', './delete-appengine.sh'
+  ]
--- a/delete-appengine.sh
+++ b/delete-appengine.sh
@@ -45,5 +45,7 @@ gcloud --project=${PROJECT_ID} \
        }
     }' | \
     while read service version ; do 
-        echo gcloud app versions delete --service $service $version
+        echo "Deleting $service $version"
+        gcloud --project=${PROJECT_ID} \
+            app versions delete --service $service $version
     done
--- a/delete-appengine.sh
+++ b/delete-appengine.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+#
+# delete-appengine.sh deletes all service versions after the first 9 with a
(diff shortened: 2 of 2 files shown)


cristinaleonr (commented):
Will the ones that get deleted always have an older `LAST_DEPLOYED` date than the 9 that are kept?

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1042#-:-Mrbf1Tm-cAEgaiB3RwU:bi9tt4n)* status: 0 of 1 approvals obtained

stephen-soltesz (commented):
Ah! very good. Dates are in the wrong order for this approach. I'll revise.

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1042#-:-MrbjX8dE2HpRyaD0iaJ:b-qmkjs1)* status: 0 of 1 approvals obtained

stephen-soltesz (commented):
@cristinaleonr PTAL? I've updated how the service versions are ordered by the list operation.

Note: there is an `echo` in place of the actual command still. This is to report what it would do in the logs without taking any action. The final step will be to remove the `echo`.

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1042#-:-Mrc8mlr1IOmGqddtMyT:b-2ny3p0)* status: 0 of 1 approvals obtained

___
*[delete-appengine.sh, line 47 at r2](https://reviewable.io/reviews/m-lab/etl/1042#-Mrc93ES9b9TSbY2hHq6:-Mrc93ES9b9TSbY2hHq7:bxw2y9f) ([raw file](https://github.com/m-lab/etl/blob/c2207bda7a4a97c6fbd82c0b1bcc36891f7f1e2a/delete-appengine.sh#L47)):*
> ```Shell
>        }
>     }' | \
>     while read service version ; do 
> ```

nit: end of line whitespace somehow...

cristinaleonr (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1042#-:-MrcFx3C4z39LuBgGmd3:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

stephen-soltesz:
It works!

PR #1044, opened 2022-01-18
Add Gardener v2 parser for switch datatype

robertodauria:
This PR replaces the current DISCO parser with a new "switch" datatype parser that uses the Sink interface and is compatible with Gardener v2.

- Output column names have been changed to match standard column names and the switch parser design document
- Row size (exported via a Prometheus metric) is calculated according to the datatype size on BQ's [Pricing](https://cloud.google.com/bigquery/pricing#storage) page
- Raw data is re-organized by the parser so that each row contains all the metrics collected by a machine for a given timestamp
- A synthetic unique ID is created for each row by concatenating the machine's hostname and the timestamp (this is required by Gardener to perform deduplication)

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/1044)


stephen-soltesz (changes requested):
Thank you!

> Reviewed 8 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1044#-:-MtjfF-W6MQDVlwsOMQm:b-9755fn)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[parser/switch.go, line 1 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtipK2V2bKlVroGvN7k:-MtipK2V2bKlVroGvN7l:b-nbehof) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L1)):*
> ```Go
> package parser
> ```

Thank you for the file rename!
___
*[parser/switch.go, line 134 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-Mtiv7CcCrjceJrDXu6G:-Mtiv7CcCrjceJrDXu6H:b-xela37) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L134)):*
> ```Go
> 				// Create the row.
> 				row = &schema.SwitchRow{
> 					ID:   fmt.Sprintf("%s%d", tmp.Hostname, sample.Timestamp),
> ```

Will users of the ID be able to reconstruct the ID easily?

Would it help to place a '-' between the name and timestamp?
___
*[parser/switch.go, line 203 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtjcMCn1OyB450J1LSQ:-MtjcMCn1OyB450J1LSR:bg6bbdf) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L203)):*
> ```Go
> func getSummaryFromSample(metric string, sample *schema.Sample, row *schema.SwitchRow) {
> 	// Convert the metric name to its corresponding CamelCase field name.
> 	val := strcase.ToCamel(metric)
> ```

Rather than `val` perhaps `delta`?
___
*[parser/switch.go, line 206 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-Mtjc2GJ-VtuFyE47b-w:-Mtjc2GJ-VtuFyE47b-x:b-5ij8c1) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L206)):*
> ```Go
> 	counter := val + "Counter"
> 
> 	// Use the "reflect" package to dinamically access the fields of the
> ```

nit: `dynamically`
___
*[parser/switch.go, line 207 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-Mtjc64h2TZG8MJjpWhd:-Mtjc64h2TZG8MJjpWhe:b-vdmbdv) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L207)):*
> ```Go
> 
> 	// Use the "reflect" package to dinamically access the fields of the
> 	// summary struct. This is a bit hacky, but it works better than a lengthy
> ```

This is not hacky :)
___
*[parser/switch.go, line 219 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtjdC7nCT5I_e8gwsQR:-MtjdC7nCT5I_e8gwsQS:bec0zbb) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L219)):*
> ```Go
> 	// Note: the octets.local.tx/rx values were not collected correctly
> 	// by DISCOv2, so we set them to zero until we can fix that.
> 	if metric == "switch.octets.local.tx" ||
> ```

This is applied to v1 and v2 data? - A we recently deployed a fix right? Is there a way to make this conditional on archive date instead so we are not truncating more data than necessary?
___
*[parser/switch.go, line 226 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtjdTxX7CvVpjsA9iSZ:-MtjdTxX7CvVpjsA9iS_:bc0ivqd) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L226)):*
> ```Go
> 	}
> 
> 	valField.SetInt(int64(sample.Value))
> ```

The value is a float64? ...  Should the summary record delta fields also be floats? Or is this intentionally truncating the values? Please document :)
___
*[parser/switch_test.go, line 64 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-Mtjed1Q5Wr9-ER9WCDL:-Mtjed1Q5Wr9-ER9WCDM:b-x31wax) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch_test.go#L64)):*
> ```Go
> 	}
> 
> 	if err := n.ParseAndInsert(meta, switchDISCOv1Filename, data); err != nil {
> ```

Since we're doing some specific things to the values read from the file, is it possible to validate those values are what we expect? e.g. "switch.octets.local.tx" is present for the v1 data but reset for the v2 data - and if you make the dates conditional verification that this logic wai - and perhaps that the summary record contains the value in the raw sample (maybe just the discard values since we care most about those?)?
___
*[schema/switch_schema.go, line 15 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-Mtip_4B8qBAuuW8pRfw:-Mtip_4B8qBAuuW8pRfx:b-ooy3ld) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/schema/switch_schema.go#L15)):*
> ```Go
> // SwitchRow represents a single row of Switch data, taken from the raw DISCO
> // exported file.
> type SwitchRow struct {
> ```

Not necessarily for this PR, but it is time to add a schema description file for this v2 type.

This will allow us to specify authoritatively how to interpret the start and end time of sample windows the meaning of this ID, etc.

https://github.com/m-lab/etl/tree/master/schema/descriptions 
___
*[schema/switch_schema.go, line 155 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-Mtis2_j7oY4h8ZuH__z:-Mtis2_j7oY4h8ZuH_a-:b-gdxskf) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/schema/switch_schema.go#L155)):*
> ```Go
> // Sample is an individual measurement taken by DISCO.
> // NOTE: the types of the fields in this struct differ from the types used
> // natively by the structs in DISCOv2. In DiSCOv2 Value is a uint64, but must
> ```

nit: tiny typo in `DiSCO`
___
*[schema/switch_schema.go, line 175 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtiuJmg7FqBCD8uBPO1:-MtiuJmg7FqBCD8uBPO2:b-mymlos) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/schema/switch_schema.go#L175)):*
> ```Go
> }
> 
> // SwitchStats represents a row of data taken from the raw DISCO export file.
> ```

Could we call this `RawSwitchStats`?

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1044)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @robertodauria and @stephen-soltesz)

___
*[parser/switch.go, line 134 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-Mtiv7CcCrjceJrDXu6G:-MtmVqFx-sWlWyFM-kIb:b9s2v4e) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L134)):*

Will users of the ID be able to reconstruct the ID easily?

Would it help to place a '-' between the name and timestamp?
</blockquote></details>

After thinking a bit more about this, `machine-site-timestamp` is probably better. It only uses fields that are available in the summary, since we decided not to store the whole hostname in there to make joining with other tables easier.
___
*[parser/switch.go, line 203 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtjcMCn1OyB450J1LSQ:-MtmW4qR5FkjsQszImTL:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L203)):*

Rather than `val` perhaps `delta`?
</blockquote></details>

Done.
___
*[parser/switch.go, line 207 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-Mtjc64h2TZG8MJjpWhd:-MtmWFWC-F3Y3umaUbZL:b-e8ih7x) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L207)):*

This is not hacky :)
</blockquote></details>

Done. This is probably a case of [_excusatio non petita, accusatio manifesta_](https://en.wikipedia.org/wiki/Excusatio_non_petita,_accusatio_manifesta) :)
___
*[parser/switch.go, line 219 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtjdC7nCT5I_e8gwsQR:-Mtma3OoB8kSl_LLDkPS:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L219)):*

This is applied to v1 and v2 data? - A we recently deployed a fix right? Is there a way to make this conditional on archive date instead so we are not truncating more data than necessary?
</blockquote></details>

Done.
___
*[parser/switch.go, line 226 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtjdTxX7CvVpjsA9iSZ:-Mtma54oAOM3vhP7gfki:b-9vd15n) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L226)):*

The value is a float64? ...  Should the summary record delta fields also be floats? Or is this intentionally truncating the values? Please document :)
</blockquote></details>

This is a float64 in the Sample struct because DISCOv1 data had floats, for some reason. To accommodate for both DISCOv1 and DISCOv2 data, I kept it float, but it's truncated to int before writing the output to the Sink (it's bytes, so I think this is correct). I've added a comment to explain why it's truncated there, while the original reason for it being a float is in a comment for the Sample struct in switch_schema.go.
___
*[schema/switch_schema.go, line 175 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtiuJmg7FqBCD8uBPO1:-Mtmax5A795SDSlTbbdz:b-896fix) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/schema/switch_schema.go#L175)):*

Could we call this `RawSwitchStats`?
</blockquote></details>

Done.

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1044)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[parser/switch_test.go, line 64 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-Mtjed1Q5Wr9-ER9WCDL:-MtnPYlmAiTpNsXKll5E:baprvba) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch_test.go#L64)):*

Since we're doing some specific things to the values read from the file, is it possible to validate those values are what we expect? e.g. "switch.octets.local.tx" is present for the v1 data but reset for the v2 data - and if you make the dates conditional verification that this logic wai - and perhaps that the summary record contains the value in the raw sample (maybe just the discard values since we care most about those?)?
</blockquote></details>

I've added some more checks. PTAL again?

stephen-soltesz (approved):
A few more comments, but I think this :lgtm:

The non-deterministic output order need not block this change but I worry about it haunting us.

> Reviewed 1 of 3 files at r3, 2 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1044#-:-MtnrC-l-xStCFqKwPGL:b-l1i2y)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @robertodauria)

___
*[parser/switch.go, line 207 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-Mtjc64h2TZG8MJjpWhd:-MtnbANrC6KC0ShtWF_R:b2nhoia) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L207)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

Done. This is probably a case of [_excusatio non petita, accusatio manifesta_](https://en.wikipedia.org/wiki/Excusatio_non_petita,_accusatio_manifesta) :)
</blockquote></details>

TIL
___
*[parser/switch.go, line 219 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtjdC7nCT5I_e8gwsQR:-MtnoVdr97HrDMO9cOFa:b-9755fa) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L219)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Thank you.
___
*[parser/switch.go, line 226 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtjdTxX7CvVpjsA9iSZ:-Mtnpe1B2fjIV-ZZhHnD:b-86rvbg) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L226)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

This is a float64 in the Sample struct because DISCOv1 data had floats, for some reason. To accommodate for both DISCOv1 and DISCOv2 data, I kept it float, but it's truncated to int before writing the output to the Sink (it's bytes, so I think this is correct). I've added a comment to explain why it's truncated there, while the original reason for it being a float is in a comment for the Sample struct in switch_schema.go.
</blockquote></details>

Please acknowledge in the comment that there is potentially some loss of information converting from discov1 to int.



_Code quote:_
```Go
	// but we want the stored values to be truncated to int.
```
___
*[parser/switch_test.go, line 51 at r5](https://reviewable.io/reviews/m-lab/etl/1044#-Mtnq554ARPVvRtbAO46:-Mtnq554ARPVvRtbAO47:bsho2ux) ([raw file](https://github.com/m-lab/etl/blob/f12bdded2035f09d7ca1a8e818fdb7f384b1a962/parser/switch_test.go#L51)):*
> ```Go
> 
> 	// Verify that the data was parsed correctly after sorting the output.
> 	sort.Slice(sink.data, func(i, j int) bool {
> ```

During our conversation I imagined the parser would sort the output. I fear that non-deterministic output from the parser could haunt us.

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1044)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[parser/switch.go, line 226 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtjdTxX7CvVpjsA9iSZ:-MtrknsK1TjLHCcUFyJC:b3cb085) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L226)):*

Please acknowledge in the comment that there is potentially some loss of information converting from discov1 to int.
</blockquote></details>

Done. I've updated the comment. It's still unclear to me how DISCOv1 could have floats different than `.0` since the delta is in bytes and the counter on the switch is in bytes, too, but I've found many examples of that in DISCOv1 data in BigQuery so there's definitely some loss of information.
___
*[parser/switch_test.go, line 51 at r5](https://reviewable.io/reviews/m-lab/etl/1044#-Mtnq554ARPVvRtbAO46:-MtrlEAyBKodkYA0Eg9Y:by3z4x6) ([raw file](https://github.com/m-lab/etl/blob/f12bdded2035f09d7ca1a8e818fdb7f384b1a962/parser/switch_test.go#L51)):*

During our conversation I imagined the parser would sort the output. I fear that non-deterministic output from the parser could haunt us.
</blockquote></details>

Done. Yes, while I suspect it won't make a difference when the data is written to BigQuery, it makes it easier to reason about the output in JSON format during local development. The output of the parser is sorted by timestamp now.

stephen-soltesz (approved):
Thank you, :lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1044#-:-Mts9fFT9VQQd3IEDrCJ:b-ses78c)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[parser/switch.go, line 226 at r2](https://reviewable.io/reviews/m-lab/etl/1044#-MtjdTxX7CvVpjsA9iSZ:-Mts8_-L09R8VloBLYcI:b-fh5p89) ([raw file](https://github.com/m-lab/etl/blob/719e5cffe48ce8964d1d9a5cd583a0a8b595d428/parser/switch.go#L226)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

Done. I've updated the comment. It's still unclear to me how DISCOv1 could have floats different than `.0` since the delta is in bytes and the counter on the switch is in bytes, too, but I've found many examples of that in DISCOv1 data in BigQuery so there's definitely some loss of information.
</blockquote></details>

Thank you for the comment.

For background, discov1 was based on collectd+RRDtool - and periodic exports from the rrd db. I think it was trying to be helpful and was calculating some 10sec average (even though the export should be raw) that resulted in the float values. It's been too long that I don't remember the details beyond that.

Issue #1052, opened 2022-02-07
scamper1 parsing errors


stephen-soltesz:
After deploying the new alternative ETL pipeline SLIs, we found that the scamper1 datatype would report parse errors after restarting:

<img width="873" alt="Screen Shot 2022-02-07 at 2 24 28 PM" src="https://user-images.githubusercontent.com/1085316/152857622-48fbf1eb-ca51-4428-ba72-c78a066fb251.png">

We suspected this may be due to a temporary format error in the early version, but have not yet confirmed.

After deploying both a modified version of the v1 parser for the "traceroute"/paris1 datatype and after copying the ndt/traceroute data to ndt/scamper1, the error rate is about 20%

<img width="922" alt="Screen Shot 2022-02-07 at 2 24 45 PM" src="https://user-images.githubusercontent.com/1085316/152858032-692968cb-18b4-4e6e-ab7b-e69d51402fb6.png">

The causes of both is currently unknown, but the parser should recognize these as either "invalid" measurements (to not include them in the set of measurements used to calculate the error rate) or to fix the parser to recognize these files.

PR #1053, opened 2022-02-08
Rename etl_test_count and update it for unsuccessful scamper parsing

cristinaleonr:
https://github.com/m-lab/etl/issues/1052 for more details.

This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/1053)


stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1053)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[parser/scamper1.go, line 118 at r1](https://reviewable.io/reviews/m-lab/etl/1053#-MvPGQdp3fWyaxi4Vrp6:-MvPGQdp3fWyaxi4Vrp7:b-i2992v) ([raw file](https://github.com/m-lab/etl/blob/729a6dd94c0d866b00560ce8d3af6e018928d225/parser/scamper1.go#L118)):*
> ```Go
> 	if err != nil {
> 		date := fileMetadata["date"].(civil.Date)
> 		if legacyScamperEnd.Before(date) {
> ```

Does this change how `TaskTotal` (etl_task_total) records the error?

If we know that this case is "invalid" we want to exclude it from the total task accounting. The basic form of the SLI equation is "good/valid events" -- our alerts are looking at "bad/valid events" but that's equivalent if we exclude invalid events. - https://www.coursera.org/lecture/site-reliability-engineering-slos/the-sli-equation-qU8h2

There may need to be an update to the worker.go also to treat some errors differently, e.g. define a base error like ErrIsInvalid then make the TaskTotal accounting conditional on `errors.Is(ErrIsInvalid)`. There may be other ways too.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1053)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[parser/scamper1.go, line 118 at r1](https://reviewable.io/reviews/m-lab/etl/1053#-MvPGQdp3fWyaxi4Vrp6:-MvVcnPN4AMQLGcTcHnK:b-9b6vil) ([raw file](https://github.com/m-lab/etl/blob/729a6dd94c0d866b00560ce8d3af6e018928d225/parser/scamper1.go#L118)):*

Does this change how `TaskTotal` (etl_task_total) records the error?

If we know that this case is "invalid" we want to exclude it from the total task accounting. The basic form of the SLI equation is "good/valid events" -- our alerts are looking at "bad/valid events" but that's equivalent if we exclude invalid events. - https://www.coursera.org/lecture/site-reliability-engineering-slos/the-sli-equation-qU8h2

There may need to be an update to the worker.go also to treat some errors differently, e.g. define a base error like ErrIsInvalid then make the TaskTotal accounting conditional on `errors.Is(ErrIsInvalid)`. There may be other ways too.
</blockquote></details>

Oh, that's a good point. I've excluded it from the `TaskTotal` metric.
___
*[task/task.go, line 160 at r2](https://reviewable.io/reviews/m-lab/etl/1053#-MvVcbOfAUdGExfCvr9f:-MvVcbOg9mHO97gdyos5:bjuikex) ([raw file](https://github.com/m-lab/etl/blob/fba3a7fae57c8b1c58adbdb281192725d1698786/task/task.go#L160)):*
> ```Go
> 		// Shouldn't have any of these, as they should be handled in ParseAndInsert.
> 		if loopErr != nil {
> 			metrics.TaskTotal.WithLabelValues(
> ```

Remove double counting of task errors. These are caught by the caller (`worker.go`).

stephen-soltesz (changes requested):
> Reviewed 1 of 3 files at r1, 3 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1053)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[parser/scamper1_test.go, line 94 at r2](https://reviewable.io/reviews/m-lab/etl/1053#-MvW8eiPA4hSQjVpsxCp:-MvW8eiPA4hSQjVpsxCq:bq1ommi) ([raw file](https://github.com/m-lab/etl/blob/bf86bebccb4f39cdeb5d9e9ced9c33936343ae6c/parser/scamper1_test.go#L94)):*
> ```Go
> 
> 			err = n.ParseAndInsert(meta, file, data)
> 			if err.Error() != tt.want.Error() {
> ```

When possible, prefer using the `errors.Is` pattern by wrapping errors. Because etl has a long legacy this may not be easy or possible. So feel free to read this note as a nice-to-have for the future. In general comparing error strings is a little more brittle.
___
*[worker/worker.go, line 229 at r2](https://reviewable.io/reviews/m-lab/etl/1053#-MvW8Ha075Ii5UanG1-E:-MvW8Ha1Eya60-yb0o6G:b4pvr2g) ([raw file](https://github.com/m-lab/etl/blob/bf86bebccb4f39cdeb5d9e9ced9c33936343ae6c/worker/worker.go#L229)):*
> ```Go
> 	if err != nil {
> 		if !errors.Is(err, parser.ErrIsInvalid) {
> 			metrics.TaskTotal.WithLabelValues(path.DataType, "TaskError").Inc()
> ```

Hmm, I'm a little confused. Does this mean that if any file inside a task archive fails to parse the entire task is recorded as an error?

And, b/c `failfast == true` in `tsk.ProcessAllTests(true)` means any bad file stops the archive from processing the rest?

Gardener has a Job (date) which produces many parser Tasks (archives) which contains many files. Before this moment I was thinking TaskTotal was counting files as "ok" or failed... like what percentage of valid files are parsing successfully... This makes more sense how the pre 2019 error rate was ~20%. I'm wondering if `ProcessAllTests` should be returning two numbers, total valid files and either files that succeeded or errored?

I'm also wondering if the SLIs based on etl_task_total are what I thought they were.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1053)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[parser/scamper1_test.go, line 94 at r2](https://reviewable.io/reviews/m-lab/etl/1053#-MvW8eiPA4hSQjVpsxCp:-MwW_-GB2N0Z2wPPQXi1:b34uopr) ([raw file](https://github.com/m-lab/etl/blob/bf86bebccb4f39cdeb5d9e9ced9c33936343ae6c/parser/scamper1_test.go#L94)):*

When possible, prefer using the `errors.Is` pattern by wrapping errors. Because etl has a long legacy this may not be easy or possible. So feel free to read this note as a nice-to-have for the future. In general comparing error strings is a little more brittle.
</blockquote></details>

This has been removed.
___
*[worker/worker.go, line 229 at r2](https://reviewable.io/reviews/m-lab/etl/1053#-MvW8Ha075Ii5UanG1-E:-MwW_IU9CClqNc2UIqyB:bsrwq4j) ([raw file](https://github.com/m-lab/etl/blob/bf86bebccb4f39cdeb5d9e9ced9c33936343ae6c/worker/worker.go#L229)):*

Hmm, I'm a little confused. Does this mean that if any file inside a task archive fails to parse the entire task is recorded as an error?

And, b/c `failfast == true` in `tsk.ProcessAllTests(true)` means any bad file stops the archive from processing the rest?

Gardener has a Job (date) which produces many parser Tasks (archives) which contains many files. Before this moment I was thinking TaskTotal was counting files as "ok" or failed... like what percentage of valid files are parsing successfully... This makes more sense how the pre 2019 error rate was ~20%. I'm wondering if `ProcessAllTests` should be returning two numbers, total valid files and either files that succeeded or errored?

I'm also wondering if the SLIs based on etl_task_total are what I thought they were.
</blockquote></details>

Right, `TaskTotal` is counting the number of tasks, not individual files, and it marks the entire task as failed if any of the files cannot be parsed. 

As for the `failfast` option, it seems to come from this change https://github.com/m-lab/etl/pull/906 and it does stop the processing when `failfast == true`.

If we want to base the failure rate on the number of individual files, we can use the `TestCount` metric instead for the SLI. I have renamed it to `TestTotal` like we did before. It also makes the proportion of errors (even pre-scamper1) small enough that we don't have to skip them https://grafana.mlab-sandbox.measurementlab.net/d/q4MrNzh7k/pipeline-slis?orgId=1&viewPanel=4&editPanel=4.

stephen-soltesz (changes requested):
> Reviewed 19 of 22 files at r4.
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1053)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr and @gfr10598)

___
*[parser/scamper1.go, line 111 at r4](https://reviewable.io/reviews/m-lab/etl/1053#-MwXew4xCOhKOYFaQcxl:-MwXew4xCOhKOYFaQcxm:bc1lo0q) ([raw file](https://github.com/m-lab/etl/blob/32ae555a248a3e3f1d80b9fb789f3bfdc5f9da91/parser/scamper1.go#L111)):*
> ```Go
> 	scamperOutput, err := parser.ParseTraceroute(rawContent)
> 	if err != nil {
> 		metrics.TestTotal.WithLabelValues(p.TableName(), scamper1, err.Error()).Inc()
> ```

What is the content of this error. If there are variable strings (e.g. filenames, pids, etc) this may be a high-cardinatliy value which is not suitable for a prometheus metric label. Every unique set of metric label values creates a unique timeseries that Prometheus must index. An explosion of these values can create higher-than-desired overhead for the Prometheus server and queries to the server.

Also, I tried to look in github.com/m-lab/traceroute-caller/parser but could not find this function. Where is it?
___
*[worker/worker.go, line 229 at r2](https://reviewable.io/reviews/m-lab/etl/1053#-MvW8Ha075Ii5UanG1-E:-MwXdlaD8P1JpGjRcR8z:b-xekpdp) ([raw file](https://github.com/m-lab/etl/blob/bf86bebccb4f39cdeb5d9e9ced9c33936343ae6c/worker/worker.go#L229)):*

Right, `TaskTotal` is counting the number of tasks, not individual files, and it marks the entire task as failed if any of the files cannot be parsed. 

As for the `failfast` option, it seems to come from this change https://github.com/m-lab/etl/pull/906 and it does stop the processing when `failfast == true`.

If we want to base the failure rate on the number of individual files, we can use the `TestCount` metric instead for the SLI. I have renamed it to `TestTotal` like we did before. It also makes the proportion of errors (even pre-scamper1) small enough that we don't have to skip them https://grafana.mlab-sandbox.measurementlab.net/d/q4MrNzh7k/pipeline-slis?orgId=1&viewPanel=4&editPanel=4.
</blockquote></details>

Okay - I have no objection to the name change.

Okay - I'm still not sure we want failfast behavior in the v2 pipeline. Especially if early scamper1 data triggers it. If only a single "bad" file is being counted per archive then the ratio may indeed be low, but it may not be representative.

@gfr10598 do you recall why you added failfast=true for the v2 pipeline?

@cristinaleonr - proposal - revert the scamper1 changes from this PR so that we can commit the name change, then resume the consideration of how to address scamper1 specifically. It doesn't make sense to me that we're not considering the known invalid date periods any more. I'm considering these things quickly so if I'm missing something please help me see.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1053)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @gfr10598 and @stephen-soltesz)

___
*[parser/scamper1.go, line 111 at r4](https://reviewable.io/reviews/m-lab/etl/1053#-MwXew4xCOhKOYFaQcxl:-MwXvMcb7WSL7_aEGMAY:b-1v64zj) ([raw file](https://github.com/m-lab/etl/blob/32ae555a248a3e3f1d80b9fb789f3bfdc5f9da91/parser/scamper1.go#L111)):*

What is the content of this error. If there are variable strings (e.g. filenames, pids, etc) this may be a high-cardinatliy value which is not suitable for a prometheus metric label. Every unique set of metric label values creates a unique timeseries that Prometheus must index. An explosion of these values can create higher-than-desired overhead for the Prometheus server and queries to the server.

Also, I tried to look in github.com/m-lab/traceroute-caller/parser but could not find this function. Where is it?
</blockquote></details>

These errors come from https://github.com/m-lab/traceroute-caller/blob/874dfadcc9fa5c2b9225f6a60e031f30eacf5229/parser/parser.go#L10.

The TRC code was refactored and the `ParseTraceroute()` function renamed `ParseRawData()`. We will have to update it when we pull the latest version of TRC.

As discussed, we will not change the metric in Prometheus at this time.
___
*[worker/worker.go, line 229 at r2](https://reviewable.io/reviews/m-lab/etl/1053#-MvW8Ha075Ii5UanG1-E:-MwXusYZEYusd94m9VEY:bdy4rla) ([raw file](https://github.com/m-lab/etl/blob/bf86bebccb4f39cdeb5d9e9ced9c33936343ae6c/worker/worker.go#L229)):*

Okay - I have no objection to the name change.

Okay - I'm still not sure we want failfast behavior in the v2 pipeline. Especially if early scamper1 data triggers it. If only a single "bad" file is being counted per archive then the ratio may indeed be low, but it may not be representative.

@gfr10598 do you recall why you added failfast=true for the v2 pipeline?

@cristinaleonr - proposal - revert the scamper1 changes from this PR so that we can commit the name change, then resume the consideration of how to address scamper1 specifically. It doesn't make sense to me that we're not considering the known invalid date periods any more. I'm considering these things quickly so if I'm missing something please help me see.
</blockquote></details>

As discussed offline, let's just rename the metric for now (and make sure to also update it for failed `scamper1` tests).

Specifically, the plan is to:

- Change the metric name (this PR).
- Export the `errTracerouteFile` error in TRC.
- Update etl to use the latest version of TRC and filter out `errTracerouteFile` errors for legacy scamper parsing.
- Investigate the `failfast` behavior.

stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1053#-:-MwY-Mpd1TH7CrYp21_j:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[worker/worker.go, line 229 at r2](https://reviewable.io/reviews/m-lab/etl/1053#-MvW8Ha075Ii5UanG1-E:-MwY-ExZ1v9wBboYL5W3:b-9755fn) ([raw file](https://github.com/m-lab/etl/blob/bf86bebccb4f39cdeb5d9e9ced9c33936343ae6c/worker/worker.go#L229)):*

As discussed offline, let's just rename the metric for now (and make sure to also update it for failed `scamper1` tests).

Specifically, the plan is to:

- Change the metric name (this PR).
- Export the `errTracerouteFile` error in TRC.
- Update etl to use the latest version of TRC and filter out `errTracerouteFile` errors for legacy scamper parsing.
- Investigate the `failfast` behavior.
</blockquote></details>

Thank you!

PR #1060, opened 2022-03-02
Only failfast for write errors

cristinaleonr:
The `failfast` option (introduced in https://github.com/m-lab/etl/pull/906) terminates a task when there is any parsing error. 

This PR makes this behavior exclusive to write errors.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/1060)


task/task.go
```diff
@@ -157,11 +159,10 @@ OUTER:
 		loopErr = tt.Parser.ParseAndInsert(tt.meta, testname, data)
 		// Shouldn't have any of these, as they should be handled in ParseAndInsert.
 		if loopErr != nil {
-			metrics.TaskTotal.WithLabelValues(
```
cristinaleonr, line 160: Avoid double counting this error. Let the caller count it.

stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1060)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[row/row.go, line 32 at r2](https://reviewable.io/reviews/m-lab/etl/1060#-MxCKsYx0jAWiZUlvEdt:-MxCKsYy7BfYUov9A1-6:b-pb5kov) ([raw file](https://github.com/m-lab/etl/blob/a91ea43950205fcc9c93cc6cc65a411ea7f04769/row/row.go#L32)):*
> ```Go
> )
> 
> // ErrCommitRow is returned when there was an error committing a
> ```

Please help me understand why we need a new error type and why the `%w` / `errors.Is` pattern is not suitable for this case. If possible that pattern would be simpler.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1060)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[row/row.go, line 32 at r2](https://reviewable.io/reviews/m-lab/etl/1060#-MxCKsYx0jAWiZUlvEdt:-MxFgV0L6ZyTcyU2OkAj:b-b9wto9) ([raw file](https://github.com/m-lab/etl/blob/a91ea43950205fcc9c93cc6cc65a411ea7f04769/row/row.go#L32)):*

Please help me understand why we need a new error type and why the `%w` / `errors.Is` pattern is not suitable for this case. If possible that pattern would be simpler.
</blockquote></details>

There is a lot of different types of errors that can be thrown when [committing](https://github.com/m-lab/etl/blob/3587755e35844668bcb9307a84ee3fb6fc498f13/storage/rowwriter.go#L87) a row. I added a general one to encapsulate them all. 

Since the base error can vary, we can't use `errors.Is` to check for value equality. So, we're using `errors.As` to check the [type](https://g3doc.corp.google.com/company/teams/go-community/gotip/episodes/32.md?cl=head).

stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1060)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[row/row.go, line 32 at r2](https://reviewable.io/reviews/m-lab/etl/1060#-MxCKsYx0jAWiZUlvEdt:-MxFtYvb81fcwbM3IbLK:b-337dtl) ([raw file](https://github.com/m-lab/etl/blob/a91ea43950205fcc9c93cc6cc65a411ea7f04769/row/row.go#L32)):*

There is a lot of different types of errors that can be thrown when [committing](https://github.com/m-lab/etl/blob/3587755e35844668bcb9307a84ee3fb6fc498f13/storage/rowwriter.go#L87) a row. I added a general one to encapsulate them all. 

Since the base error can vary, we can't use `errors.Is` to check for value equality. So, we're using `errors.As` to check the [type](https://g3doc.corp.google.com/company/teams/go-community/gotip/episodes/32.md?cl=head).
</blockquote></details>

Do we introspect the base errors (do they need to be preserved?)

For example, if we defined a standard error type for `ErrCommitRow` and combined the returned error with our error, we could use `errors.Is` later.

So, row.go could return an error like:

```
var ErrCommitRow = errors.New("failed to commit row")
...

func (pb *Base) commit(rows []interface{}) error {
    ...
	if err != nil {
	    log.Println(pb.label, err)
            pb.stats.Done(len(rows)-done, err)
            return fmt.Errorf("%v: %w", err, ErrCommitRow)
        }
        return nil
}
```

If this is not right, no problem; please say more.

stephen-soltesz (changes requested):
Wait I didn't mean to approve it -- sorry want to be sure first.

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1060#-:-MxFvlTI2fm1tNAm3qJF:bonw7ir)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1060)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[row/row.go, line 32 at r2](https://reviewable.io/reviews/m-lab/etl/1060#-MxCKsYx0jAWiZUlvEdt:-MxG4S5V5FCgaNXmOPqH:b-r5met) ([raw file](https://github.com/m-lab/etl/blob/a91ea43950205fcc9c93cc6cc65a411ea7f04769/row/row.go#L32)):*

Do we introspect the base errors (do they need to be preserved?)

For example, if we defined a standard error type for `ErrCommitRow` and combined the returned error with our error, we could use `errors.Is` later.

So, row.go could return an error like:

```
var ErrCommitRow = errors.New("failed to commit row")
...

func (pb *Base) commit(rows []interface{}) error {
    ...
	if err != nil {
	    log.Println(pb.label, err)
            pb.stats.Done(len(rows)-done, err)
            return fmt.Errorf("%v: %w", err, ErrCommitRow)
        }
        return nil
}
```

If this is not right, no problem; please say more.
</blockquote></details>

We do check the base error type. For example, [here](https://github.com/m-lab/etl/blob/3587755e35844668bcb9307a84ee3fb6fc498f13/task/task.go#L185) (I've also improved this check :).

In my opinion, preserving them is a better practice.

stephen-soltesz (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1060)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[row/row.go, line 32 at r2](https://reviewable.io/reviews/m-lab/etl/1060#-MxCKsYx0jAWiZUlvEdt:-MxG8RI3CMWYspgXSZIJ:b-vf0dfy) ([raw file](https://github.com/m-lab/etl/blob/a91ea43950205fcc9c93cc6cc65a411ea7f04769/row/row.go#L32)):*

We do check the base error type. For example, [here](https://github.com/m-lab/etl/blob/3587755e35844668bcb9307a84ee3fb6fc498f13/task/task.go#L185) (I've also improved this check :).

In my opinion, preserving them is a better practice.
</blockquote></details>

Thank you!

Until now, I had not seen a use-case where actually defining a custom error type was preferable. What I'm taking from this is that in middleware packages, it may be necessary to wrap errors this way. I would continue to advocate for base packages defining their own simple errors so third party users can easily identify them using `errors.Is`.

stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1060#-:-MxGMWuT1Xetm1XXjrEy:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #1105, opened 2022-10-10
Fix thinSnaps

stephen-soltesz:
This change completes fixes the bug reported by @NotSpecial - https://github.com/m-lab/etl/pull/1104 with an update to the unit test to check that every `FinalSnapshot` matches the last snapshot of the raw, thinned snapshots.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/1105)

### DIFF
--- a/parser/tcpinfo.go
+++ b/parser/tcpinfo.go
@@ -96,15 +96,19 @@ func (p *TCPInfoParser) IsParsable(testName string, data []byte) (string, bool) 
 	return "", false
 }
 
+// thinSnaps collects every 10th snapshot in orig. thinSnaps guarantees that the last snapshot is preserved.
 func thinSnaps(orig []snapshot.Snapshot) []snapshot.Snapshot {
 	n := len(orig)
+	if n == 0 {
+		return orig
+	}
 	out := make([]snapshot.Snapshot, 0, 1+n/10)
-	for i := 0; i < n; i += 10 {
+	// Take first and every 10th snapshot after (exluding the final snapshot, which is always included).
+	for i := 0; i < n-1; i += 10 {
 		out = append(out, orig[i])
 	}
-	if n%10 != 0 {
-		out = append(out, orig[n-1])
-	}
+	// Always append final snapshot.
+	out = append(out, orig[n-1])
 	return out
 }
 
--- a/parser/tcpinfo_test.go
+++ b/parser/tcpinfo_test.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/civil"
+	"github.com/go-test/deep"
 	"github.com/m-lab/etl/etl"
 	"github.com/m-lab/etl/parser"
 	"github.com/m-lab/etl/schema"
@@ -169,6 +170,10 @@ func TestTCPParser(t *testing.T) {
 		if row.A.SockID.SrcIP != "195.89.146.242" && row.A.SockID.SrcIP != "2001:5012:100:24::242" {
 			t.Error("Wrong SrcIP", row.A.SockID.SrcIP)
 		}
+		// Guarantee that FinalSnapshot matches the last raw snapshot.
+		if diff := deep.Equal(row.A.FinalSnapshot, row.Raw.Snapshots[len(row.Raw.Snapshots)-1]); diff != nil {
+			t.Errorf("TestTCPParser.ParseAndInsert() FinalSnapshot and last snapshot differ: %s", strings.Join(diff, "\n"))
+		}
 	}
 
 	// This section is just for understanding how big these objects typically are, and what kind of compression
@@ -200,10 +205,6 @@ func TestTCPParser(t *testing.T) {
 
 	if duration > 20*time.Second {
 		t.Error("Incorrect duration calculation", duration)
-	}
-
-	if totalSnaps != 1588 {
-		t.Error("expected 1588 (thinned) snapshots, got", totalSnaps)
 	}
 }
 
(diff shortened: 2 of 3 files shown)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1105#-:-NE1vQUTChgkHE1L20sd:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained

parser/tcpinfo_test.go
```diff
@@ -169,6 +170,10 @@ func TestTCPParser(t *testing.T) {
 		if row.A.SockID.SrcIP != "195.89.146.242" && row.A.SockID.SrcIP != "2001:5012:100:24::242" {
 			t.Error("Wrong SrcIP", row.A.SockID.SrcIP)
 		}
+		// Guarantee that FinalSnapshot matches the last raw snapshot.
+		if diff := deep.Equal(row.A.FinalSnapshot, row.Raw.Snapshots[len(row.Raw.Snapshots)-1]); diff != nil {
+			t.Errorf("TestTCPParser.ParseAndInsert() FinalSnapshot and last snapshot differ: %s", strings.Join(diff, "\n"))
+		}
```
NotSpecial, line 176: Ah, nice, I think this is a very good way to test this. I had one additional thought:

Maybe we should check that the number of thinned snapshots is correct (per row, not in summary as before)?

If it helps, the expected number of thinned snapshots is: `2 + math.floor((len(row.Raw.Snapshots) - 2) / 10)`, i.e. the first and last element are always included, and every tenth other element.

N.B.: I am not 100% sure how thinSnaps behaves when raw.Snapshots is only 1 element long. Will this element be duplicated? It should not happen, but I am not 100% sure.

parser/tcpinfo.go
```diff
@@ -96,15 +96,19 @@ func (p *TCPInfoParser) IsParsable(testName string, data []byte) (string, bool)
 	return "", false
 }
 
+// thinSnaps collects every 10th snapshot in orig. thinSnaps guarantees that the last snapshot is preserved.
 func thinSnaps(orig []snapshot.Snapshot) []snapshot.Snapshot {
 	n := len(orig)
+	if n == 0 {
```
NotSpecial, line 102: Regarding [this comment](https://github.com/m-lab/etl/pull/1105/commits/bebb994b3240cf86991dc51815159bd398a60921#r991514476): I am not 100% sure how thinSnaps behaves when the list has only one element.

Theoretically, the for loop should not do anything, as for `i=0` and `n=1` the condition `i < n-1` is false, but maybe it's worth checking?

To be super safe, one could change this check to `if n<=2`. I.e. the first and last element are always included (or the single element, if there is only one); and only if we have more than two elements, we apply the thinning logic.

stephen-soltesz (commented):
I've added an explicit test for the `thinSnaps` function with edge case conditions. I like your first implementation (handling the last element unconditionally for all lengths). The tests should give us much more confidence that this behavior is correct.

The same test cases (correctly) fail for the original (buggy) implementation of `thinSnaps`.

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1105#-:-NE2FZ5p75hdiZn4R1tr:bupa52t)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @NotSpecial)

___
*[`parser/tcpinfo_test.go` line 176 at r1](https://reviewable.io/reviews/m-lab/etl/1105#-NE1ntdemUBgPCgmdCoB-r1-176:-NE2F1om4gJE_3DxykqH:b-2q9xdb) ([raw file](https://github.com/m-lab/etl/blob/bebb994b3240cf86991dc51815159bd398a60921/parser/tcpinfo_test.go#L176)):*
<details><summary><i>Previously, NotSpecial (Alexander Dietmüller) wrote…</i></summary><blockquote>

Ah, nice, I think this is a very good way to test this. I had one additional thought:

Maybe we should check that the number of thinned snapshots is correct (per row, not in summary as before)?

If it helps, the expected number of thinned snapshots is: `2 + math.floor((len(row.Raw.Snapshots) - 2) / 10)`, i.e. the first and last element are always included, and every tenth other element.

N.B.: I am not 100% sure how thinSnaps behaves when raw.Snapshots is only 1 element long. Will this element be duplicated? It should not happen, but I am not 100% sure.




</blockquote></details>

I like this idea. Unfortunately, this information (true snapshot length) is not available to the unit test currently. Though this feels like an oversight for data usage also. I can imagine the `A` record could include some additional summary information like `TotalSnaps` or `SnapCount` or `SnapshotLength` or similar to indicate that the `len(raw.Snapshots)` is not expected to be equal to the total snapshots due to the thinning... Today there is nothing to indicate this explicitly. But, this changes the scope of this change from a bug fix to a schema change... 

Would you accept an open issue for more metadata to be added in the future?

NotSpecial (approved):
Looking good to me, the explicit test is useful!

stephen-soltesz:
The parser updated with this fix was deployed to staging around 2022-10-18 20:20:00 - parse times after that have FinalSnapshot and last Snapshots with matching timestamps (`notmatching == 0`).

```
SELECT
  date,
  COUNT(*) as total,
  COUNTIF(a.FinalSnapshot.Timestamp != raw.Snapshots[SAFE_ORDINAL(ARRAY_LENGTH(raw.Snapshots))].Timestamp) as notmatching
FROM mlab-staging.ndt.tcpinfo
WHERE
  date between "2021-04-02" AND "2021-05-10"
  and parser.Time > TIMESTAMP("2022-10-18 00:00:00")
GROUP BY date
ORDER BY date 
```

After deployment to production the historical reprocessing will take about 16 days to cover all dates. Daily data will be update daily.

PR #1108, opened 2022-12-13
Add capacity to filter select client_name values

stephen-soltesz:
This change adds support for filtering select client_name values from the ndt7 ClientMetadata. This change prevents the value from being published to BigQuery, making the value effectively unsearchable, though it will remain in the underlying raw storage. Since clients optionally provide the `client_name` value, it is impossible to distinguish between clients that never provided the value or those that have the value filtered without looking at the raw data directly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/1108)


stephen-soltesz:
Obsolete.

PR #1113, opened 2023-04-05
Support more options for new experiments

stephen-soltesz:
Previously, the `update-schema` utility only supported legacy datatypes and `ndt` sidecars. The command assumed that datasets already existed. And the logic was straight forward but very redundant.

This change adds support to `update-schema` for creating datasets and sidecar tables for alternate experiments. Previous behavior is still available but no longer run by default. Now, the `-standard` and `-legacy` options are independent. We typically do not need to update the legacy tables any longer and the template table iteration is time consuming.

Example usage now includes:

```sh
# Make standard supported tables.
update-schema -project mlab-sandbox -standard

# Make all datasets and sidecar tables for new foobar experiments.
update-schema -project mlab-sandbox -experiment foobar -sidecars

# Make a single scamper1 table for the wehe experiment.
update-schema -project mlab-sandbox -experiment wehe -datatype scamper1
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/1113)

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -93,7 +93,7 @@ steps:
   id: "Update table schemas before deploying parsers"
   entrypoint: bash
   args: [
-     '-c', './update-schema', '-standard'
+     '-c', './update-schema -standard'
   ]
   env:
   - PROJECT=$PROJECT_ID
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -98,6 +98,17 @@ steps:
   env:
   - PROJECT=$PROJECT_ID
 
+- name: etl-testing
+  id: "Update table schemas for wehe deploying parsers"
+  entrypoint: bash
+  args: [
+     '-c', './update-schema -experiment wehe -datatype scamper1 &&
+            ./update-schema -experiment wehe -datatype annotation2 &&
+            ./update-schema -experiment wehe -datatype hopannotation2'
+  ]
+  env:
+  - PROJECT=$PROJECT_ID
+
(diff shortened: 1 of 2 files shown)


cristinaleonr (commented):
I know this PR is not starting it, but my opinion is that this command is doing way too many things.

Ideally, only the `main()` function would be here, and the rest would be in a separate package with unit tests.

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1113#-:-NSMjo8g2xcH_54KPqBI:b-wv0lf1)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`cmd/update-schema/update.go` line 190 at r2](https://reviewable.io/reviews/m-lab/etl/1113#-NSMjPJAD0wjzFcmQ2qL:-NSMjB6x470135TeZWSz:bs4pmyb) ([raw file](https://github.com/m-lab/etl/blob/aa0bb648cea6eeb7410a8dabada191c359b0e8b7/cmd/update-schema/update.go#L190)):*
> ```Go
> 		if !ok {
> 			log.Printf("failed to find %v", table)
> 			return 1
> ```

Should this increase the error count and continue?

cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1113#-:-NSMpuH_5sKwco2TMyeU:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained

stephen-soltesz:
Thank you!

stephen-soltesz:
Part of:
* https://github.com/m-lab/etl/issues/1114

PR #1133, opened 2024-10-22
Prevent scamper1 rows that are too large

stephen-soltesz:
Periodically, daily jobs fail due to errors like those below. Sometimes scamper1 generates rows that are greater than 100MB when the maximum supported [Load row size is 100MB](https://cloud.google.com/bigquery/quotas#load_jobs), leading to these failures. As well, since the parser/gardner system does not retry failed daily jobs, one large file prevents the whole day from loading successfully.

This change simply filters raw traces larger than 90MB to prevent this type of failure. It is an open question why scamper would ever generate a trace this large. This is likely to be an anomalous measurement already. This change does not address the root cause.

```
20241018:ndt/scamper1 Load {Location: "gs://etl-mlab-oti-us-central1/archive-measurement-lab/ndt/scamper1/2024/10/18/20241018T140116.143544Z-scamper1-mlab1-gru03-ndt.tgz.jsonl"; 
Message: "Error while reading data, error message: JSON parsing error in row starting at position  116704618:
Row size is larger than: 104857600. File: gs://etl-mlab-oti-us-central1/archive-measurement-lab/ndt/scamper1/2024/10/18/20241018T140116.143544Z-scamper1-mlab1-gru03-ndt.tgz.jsonl"; Reason: "invalid"}
```

See also: https://github.com/m-lab/dev-tracker/issues/744

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/1133)

### DIFF
--- a/parser/scamper1.go
+++ b/parser/scamper1.go
@@ -18,6 +18,10 @@ import (
 
 const (
 	scamper1 = "scamper1"
+
+	// BigQuery rows must be under 100MB. Use 90MB to allow for JSON export.
+	// This is already an unusually large trace file.
+	maxRowSize = 90000000
 )
 
 // Scamper1Parser handles parsing for the scamper1 datatype.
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/etl/1133#-:-O9oxZybEj_Dapl5Bqbd:bg7igkv)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #1135, opened 2026-07-03
fix(schema): bump ndt-server to v0.24.1 to add ConnectionInfo.StartTime

robertodauria:
## Problem

ndt-server v0.24.0 added `StartTime` to `ConnectionInfo` (m-lab/ndt-server#413). The autojoin fleet's autoloaded tables gained this field on 2026-07-01/02 via the jostler-uploaded table schemas, but this repo still pins ndt-server v0.20.18, so the parser-generated ndt7 schema (and therefore `raw_ndt.ndt7` / `ndt.ndt7`) lacks it.

Because `SELECT *` UNION ALL views require identical struct shapes, `measurement-lab.ndt.ndt7_union` (legacy fleet UNION autojoin fleet) currently fails to compile:

```
Column 6 in UNION ALL has incompatible types: STRUCT<...>, STRUCT<...>
```

## Fix

Bump ndt-server to **v0.24.1** — the newest tag that includes `ConnectionInfo.StartTime` while keeping `go 1.20` (v0.25.x requires go 1.25, which is a larger toolchain/CI change best done separately).

Verified locally:
- generated `NDT7ResultRow` schema gains exactly the four `raw.{Upload,Download}.{Client,Server}Measurements.ConnectionInfo.StartTime` TIMESTAMP fields, appended at the end of the struct (matching the position BigQuery used when the autojoin tables were updated) — no other schema changes vs the live `mlab-oti.ndt.ndt7` table
- `go build ./...` and the full test suite pass

## Propagation

On deploy, `update-schema -standard` adds the fields to `tmp_ndt.ndt7`/`raw_ndt.ndt7`; Gardener's join jobs run with `ALLOW_FIELD_ADDITION`, so `ndt.ndt7` picks the fields up on the next materialization and `ndt7_union` compiles again. The parser also starts recording `StartTime` values instead of dropping them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/etl/1135)


Dockerfile.testing
```diff
@@ -17,7 +17,7 @@ WORKDIR $GOPATH
 
 # Install binaries needed for builds and testing.
 RUN apt-get --allow-releaseinfo-change update
-RUN apt-get install -y jq gcc netcat
+RUN apt-get install -y jq gcc netcat-openbsd
```
bassosimone, line 20: ❤️ (OpenBSD's netcat is *so* good)

bassosimone (approved):
I understand the rationale of this pull request and I approve it, even though I did not have time to test the changes myself. Thank you for chasing this issue! 🙌

